PR #43, opened 2019-08-02
Improve precision of mathematical operations

simonvbrae:
Used the library decimal.js to do mathematical operations for us.

Notes:
* Option to avoid Decimal operation has not been implemented yet.
* I may have missed some operations.
This PR changes the following implementations from the RegularFunctions file: multiplication, division, addition, substraction.
* Could benefit from a more efficient implementation. Maybe using Decimals internally?

Fixes SUM test from https://github.com/comunica/comunica/issues/287

### DIFF
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -1,5 +1,6 @@
 import * as RDFDM from '@rdfjs/data-model';
 import * as hash from 'create-hash';
+import * as decimaljs from 'decimal.js';
 import * as uuid from 'uuid';
 
 import { Map } from 'immutable';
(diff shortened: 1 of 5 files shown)


lib/functions/RegularFunctions.ts
```diff
@@ -52,36 +53,41 @@ const unaryMinus = {
 const multiplication = {
   arity: 2,
   overloads: declare()
-    .arithmetic((left, right) => left * right)
+  // @ts-ignore
```
rubensworks: Are these necessary?

simonvbrae:
Notes:
* I may have missed some operations. This PR only changes the following functions from the RegularFunctions file: multiplication, division, addition, substraction.
* Typescript typings aren't added yet, hence the `@ts-ignore`'s
* I think this can be implemented more efficiently by rewriting it so the library uses Decimal instances internally, instead of number

lib/functions/RegularFunctions.ts
simonvbrae: I haven't added TS typings yet but I'm going to look into it now!

simonvbrae: TS gives the error message that .mul, or any other function, is not defined in the module.

rubensworks: That's weird. Is it possible that you're using the typings incorrectly? Perhaps try importing like this: `import {mul} from 'decimal.js';`

rubensworks:
Created #44 instead, as I don't have push access to @simonvbrae's fork.

PR #48, opened 2019-10-11
Support wildcards

joachimvh:
This makes sure sparqlee can handle algebra with wildcards in it.

This is required for comunica/comunica#542

This can only happen in 1 specific instance and that is a `count(*)`. That is the only time a wildcard can appear (yay for edge cases). The fix is to convert the wildcard to a namednode('*') which is how wildcards were previously handled. You could also make internal representations of wildcards but that would probably require quite some additional changes (and sparqlee is never going to output a wildcard).


wschella:
This should work, since the Count aggregator state machine actually ignores the term that results from evaluating the expression (the NamedNode) in this case. Should probably be tested somewhere tho, so i made #50 .

rubensworks:
Released as 1.3.0. @joachimvh, could you update the Comunica PR with this?

PR #69, opened 2020-07-08
Passing a term to sparqlee to order

stephaniech97:


lib/Transformation.ts
```diff
@@ -197,3 +197,24 @@ export function transformAggregate(expr: Alg.AggregateExpression) {
 export function transformExistence(expr: Alg.ExistenceExpression) {
   return new E.Existence(expr);
 }
+
+// Ordering 2 given strings
+export function orderTypes(litA: RDF.NamedNode | RDF.BlankNode | RDF.Literal | RDF.Variable | RDF.DefaultGraph | undefined,
```
rubensworks, line 202: `RDF.NamedNode | RDF.BlankNode | RDF.Literal | RDF.Variable | RDF.DefaultGraph` can be simplified to `RDF.Term`.

Also, I would move these two new functions into a new file (`Ordering.ts`), as this is not really a transformation.

lib/Transformation.ts
```diff
@@ -197,3 +197,24 @@ export function transformAggregate(expr: Alg.AggregateExpression) {
...
+
+// Ordering 2 given strings
+export function orderTypes(litA: RDF.NamedNode | RDF.BlankNode | RDF.Literal | RDF.Variable | RDF.DefaultGraph | undefined, 
+                        litB: RDF.NamedNode | RDF.BlankNode | RDF.Literal | RDF.Variable | RDF.DefaultGraph | undefined, 
+                        isAscending:boolean){
+  if (litA && litA.termType === "Literal" && litB && litB.termType === "Literal" ){
+    const a = transformLiteral(litA);
+    const b = transformLiteral(litB); 
+    return order(a, b, isAscending);
+  }else{
```
rubensworks, line 209: Missing spaces around `else`.

lib/Transformation.ts
```diff
@@ -197,3 +197,24 @@ export function transformAggregate(expr: Alg.AggregateExpression) {
 export function transformExistence(expr: Alg.ExistenceExpression) {
   return new E.Existence(expr);
 }
+
+// Ordering 2 given strings
+export function orderTypes(litA: RDF.NamedNode | RDF.BlankNode | RDF.Literal | RDF.Variable | RDF.DefaultGraph | undefined, 
+                        litB: RDF.NamedNode | RDF.BlankNode | RDF.Literal | RDF.Variable | RDF.DefaultGraph | undefined, 
+                        isAscending:boolean){
+  if (litA && litA.termType === "Literal" && litB && litB.termType === "Literal" ){
```
rubensworks, line 205: We typically write our strings with single quotes (our linter is probably not configured strictly enough...).

lib/util/Consts.ts
```diff
@@ -1,6 +1,6 @@
 // tslint:disable:variable-name
 import * as RDFDM from '@rdfjs/data-model';
-import { Map, Set } from 'immutable';
+import { Map, Set, is } from 'immutable';
```
rubensworks, line 3: Unneeded import?

lib/Transformation.ts
rubensworks: I'm also wondering now if literals can really be undefined. What happens if you remove these undefined-checks?

stephaniech97: const transform = async (bindings: Bindings, next: any) => {
        try {
          const result = await evaluator.evaluate(bindings);
          transformedStream._push({ bindings, result });
          
        } catch (err) {
          if (!isExpressionError(err)) {
            bindingsStream.emit('error', err);
          }
          transformedStream._push({ bindings, result: undefined });
        }
        next();
      };
In the ActorQueryOperationOrderBySparqlee there is a possibility of having an error. If for example I would pass a non-literal into the evaluate it will throw an error and then still do the ordering but with undefined.

lib/util/Consts.ts
stephaniech97: The import was never used

lib/Transformation.ts
```diff
@@ -110,7 +110,6 @@ export function transformLiteral(lit: RDF.Literal): E.Literal<any> {
 
     case DT.XSD_INTEGER:
     case DT.XSD_DECIMAL:
-
```
rubensworks, line 113: Could you undo the changes in this file? That will make the diff a bit cleaner :-)

test/util.test/Parsing.test.ts
```diff
@@ -54,4 +54,4 @@ describe('util/parsing', () => {
       });
     });
   });
-});
+});
```
rubensworks, line 57: Changes here can also be undone.

Also, it looks like your editor removes empty lines at the end of files, while these are typically good to have. Maybe a config option somewhere in your editor?

lib/util/Ordering.ts
```diff
@@ -0,0 +1,22 @@
...
+        const b = transformLiteral(litB); 
+        return order(a, b, isAscending);
+    } else {
+        return 0;
+    }
+}
+
+// Effective ordering
+export function order(orderA: T.Literal<any>, orderB: T.Literal<any>, isAscending:boolean){
+    if (!orderA || !orderB || orderA.typedValue === orderB.typedValue) {
```
rubensworks, line 18: As far as I can see, the output of `transformLiteral` is always defined. So I think these falsy checks can be removed, right?

lib/util/Ordering.ts
```diff
@@ -0,0 +1,22 @@
+import * as RDF from 'rdf-js';
+import * as T from '../expressions/Term';
+import { transformLiteral } from '../Transformation';
+
+// Ordering 2 given strings
```
rubensworks, line 5:
```suggestion
// Determine the relative numerical order of the two given terms.
```

rubensworks:
Great work, thanks!

PR #71, opened 2020-07-17
Fix bug min/max of non-numerical type, Closes #673

stephaniech97:
https://github.com/comunica/comunica/issues/673


rubensworks:
Did you check what the intended behaviour in the SPARQL spec is? Does this PR conform to this?

stephaniech97:
https://en.wikibooks.org/wiki/SPARQL/Expressions_and_Functions#COUNT,_MIN,_MAX,_AVG_and_SUM Here was mentioned that "The functions COUNT, MIN, MAX, AVG, and SUM can only be used as Aggregate functions." So when further going into the aggregate functions "https://en.wikibooks.org/wiki/SPARQL/Aggregate_functions" they defined it in the following way "MIN, MAX: the minimum or maximum value of all elements, respectively. This works for all value types; numbers are sorted numerically, strings and other types lexically." This is why I chose to accept all types. If different types however are compared, i chose to throw an error.

rubensworks:
> If different types however are compared, i chose to throw an error.

The "chose" in this sentence makes me suspect that it's not spec-compliant.
Can you find any mentions in the spec on what should happen when different types are compared?
Maybe you can try with other SPARQL engines to see what happens there?

lib/evaluators/AggregateEvaluator.ts
```diff
@@ -313,3 +312,51 @@ function extractNumericValueAndTypeOrError(term: RDF.Term): { value: number, typ
   const value = parseXSDFloat(term.value);
   return { type, value };
 }
+
+function extractValue(term: RDF.Term): {value: any, type:string}  {
+  if (term.termType !== 'Literal') {
+    throw new Error('Term is not a literal');
```
rubensworks, line 318: Could you include the value and termType in the error message? (this will make debugging easier should this error occur)

lib/evaluators/AggregateEvaluator.ts
```diff
@@ -313,3 +312,51 @@ function extractNumericValueAndTypeOrError(term: RDF.Term): { value: number, typ
...
+    if (extractedElem.type !== type) {
+      sameType = false;
+      break;
+    }
+  }
+
+  // Different types will be compared as strings
+  let extreme = state[0].term;
+  let extremeValue;
+  if (sameType) {
```
rubensworks, line 340: `value` is always a string, right? So this if-else statement should not be needed.

lib/evaluators/AggregateEvaluator.ts
```diff
@@ -313,3 +312,51 @@ function extractNumericValueAndTypeOrError(term: RDF.Term): { value: number, typ
...
+    extremeValue = extreme.value;
+  } else {
+    extremeValue = '' + extreme.value;
+  }
+
+  // Compare elements
+  let elementValue;
+  for (const element of state) {
+    if (sameType) {
```
rubensworks, line 349: Same as above.

lib/evaluators/AggregateEvaluator.ts
```diff
@@ -313,3 +312,51 @@ function extractNumericValueAndTypeOrError(term: RDF.Term): { value: number, typ
...
+      elementValue = element.value;
+    } else {
+      elementValue = '' + element.value;
+    }
+
+    if (isMin && elementValue < extremeValue) {
```
rubensworks, line 355: These two cases do the same. Can they be combined?

lib/evaluators/AggregateEvaluator.ts
stephaniech97: The term is retrieved from extractValue(term: RDF.Term). Which gets the value by using transformLiteral(term).value. This function returns values from types different from string such as dates, numericalLiterals... https://github.com/comunica/sparqlee/blob/master/lib/Transformation.ts#L73 So I don't think it is always a string in this case.

lib/evaluators/AggregateEvaluator.ts
```diff
@@ -306,10 +306,47 @@ export const aggregators: Readonly<{ [key in SetFunction]: AggregatorClass }> =
 function extractNumericValueAndTypeOrError(term: RDF.Term): { value: number, type: C.NumericTypeURL } {
   // TODO: Check behaviour
   if (term.termType !== 'Literal' || !C.NumericTypeURLs.contains(term.datatype.value)) {
-    throw new Error('Term is not numeric');
+    throw new Error('Term with value ' + term.value + ' has type ' + term.termType +' and is not a numeric literal');
```
rubensworks, line 309: If it's a literal (most of the cases), I'd also include the datatype here.

lib/evaluators/AggregateEvaluator.ts
```diff
@@ -178,49 +182,45 @@ class Sum extends BaseAggregator<SumState> {
   }
 }
 
-type MinState = { minNum: number, minTerm: RDF.Term };
-class Min extends BaseAggregator<MinState> {
-  init(start: RDF.Term): MinState {
-    const { value } = extractNumericValueAndTypeOrError(start);
-    return { minNum: value, minTerm: start };
+type ExtremeState = { value: number, term: RDF.Term };
+class Min extends BaseAggregator<ExtremeState[]> {
```
rubensworks, line 186: I didn't notice this before, but this change will make it so that this aggregator will keep _all_ passed terms in memory, which will lead to major performance problems, since the number of intermediary terms can be extremely high.

As such, I would keep the previous approach where only a singular state is maintained, instead of storing _all_ states.

Of course, this complicates the approach we wanted to take here (where string-based comparison is done when we have different types), so perhaps we need to reconsider that choice.

Perhaps we can just attempt a `parseInt` on non-numbers, and if we have a`NaN`, we simply ignore the value. What do you think?

(in this case, make sure to add a test where there are only string terms that result in `NaN`)

test/evaluators/AggregateEvaluator.test.ts
```diff
@@ -33,6 +33,10 @@ function makeAggregate(aggregator: string, distinct = false, separator?: string)
   };
 }
 
+function nonLiteral(): RDF.Term {
+  return namedNode('http://www.w3.org/2001/XMLSchema#integer');
```
rubensworks, line 37: This is confusing :P I would just use something like `http://example.org/` instead.

test/evaluators/AggregateEvaluator.test.ts
```diff
@@ -315,38 +350,73 @@ describe('an aggregate evaluator should be able to', () => {
...
-    describe('when we ask for throwing errors', () => {
-      it('and the input is empty', () => {
-        expect(() => {
-          testCase({
-            expr: makeAggregate('max'),
-            input: [],
-            throwError: true,
-          });
-        }).toThrow();
+    it('different types compare by string (min)', () => {
```
rubensworks, line 353: Test name does not correspond to what the test does. (also for the tests below)

test/evaluators/AggregateEvaluator.test.ts
```diff
@@ -315,38 +350,73 @@ describe('an aggregate evaluator should be able to', () => {
...
+          Bindings({ '?x': nonLiteral() }),
+          Bindings({ '?x': int('2') }),
+          Bindings({ '?x': int('3') }),
+          Bindings({ '?x': int('4') }),
+        ],
       });
+      expect(undefined);
+    });
 
-      it('and any value in the stream errors', () => {
```
rubensworks, line 342: Why was this test changed? The old version of this tests should still apply, right?

PR #78, opened 2020-10-31
Replaces create-hash with hash.js and spark-md5

jacoscaz:
This PR replaces `create-hash` with the combination of  `hash.js` and `spark-md5`. 

The idea came from https://github.com/comunica/comunica/pull/752 , in which @rubensworks and I discussed strategies to decrease the amount of unused crypto-related code ending up in browser-side bundles of `comunica`.


package.json
```diff
@@ -55,6 +55,8 @@
     "@types/benchmark": "^1.0.30",
     "@types/jest": "^26.0.0",
     "@types/rdf-js": "^4.0.0",
+    "@types/spark-md5": "^3.0.2",
```
rubensworks, line 58: In Comunica-related projects we have the convention to add `@types/*` for main dependencies as actual dependency (not dev), could you move these two to below?

package.json
```diff
@@ -68,15 +70,13 @@
     "typescript": "^4.0.2"
   },
   "dependencies": {
-    "@types/create-hash": "^1.2.0",
-    "@types/rdf-js": "*",
```
rubensworks, line 72: This one can remain :-)

rubensworks (changes requested):
That looks simple enough :-)
Just some package.json comments.

package.json
jacoscaz: ah, my bad, I thought these had ended there by mistake

rubensworks: FYI, this is why: https://github.com/comunica/comunica/issues/294

rubensworks:
Thanks!
I'll make sure to include this change in the new Comunica release.

PR #81, opened 2021-03-26
Preserve original blank nodes when using custom bnode generator functions

jacoscaz:
While working on https://github.com/comunica/comunica/issues/795 I realized that using custom `bnode` generator functions (also see https://github.com/comunica/sparqlee/pull/80) does not result in the generated blank nodes being returned by `sparqlee`'s evaluators. 

This PR changes how the `BlankNode` expression works internally, supporting the preservation of the blank node being wrapped into the expression itself and ultimately unwrapped by calling `.toRDF()`.

### DIFF
--- a/lib/expressions/Term.ts
+++ b/lib/expressions/Term.ts
@@ -43,11 +43,12 @@ export class NamedNode extends Term {
 export class BlankNode extends Term {
   static _nextID = 0;
 
+  value: RDF.BlankNode;
   termType: TermType = 'blankNode';
 
-  constructor(public value: string) {
+  constructor(value: RDF.BlankNode|string) {
     super();
-    this.value = value;
+    this.value = typeof value === 'string' ? DF.blankNode(value) : value;
   }
 
   static nextID() {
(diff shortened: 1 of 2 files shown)


rubensworks:
Might make sense to get rid of the string-based constructor completely. But that might require some refactoring, so let's keep that for future work :-)

jacoscaz:
Thank you for merging @rubensworks ! Any chance you could get version `1.6.2` out? Hopefully this is the last one. I did test locally by `yarn link`ing `sparqlee` into `comunica` but you never know...

rubensworks:
> Any chance you could get version 1.6.2 out?

Sure, done!

wschella:
> Might make sense to get rid of the string-based constructor completely. But that might require some refactoring, so let's keep that for future work :-)

It would force a custom BNODE generator to always be provided. Currently that's not necesary, as sparqlee provides a UUID or bnode-input-based fallback that should be compliant. 
https://github.com/comunica/sparqlee/blob/c843192619217bdfd9a23be836d4941a212d3d04/lib/functions/SpecialFunctions.ts#L366

PR #85, opened 2021-07-26
Feature - Support custom functions (closes #16)

### LINKED ISSUE #16: Support custom functions
See https://www.w3.org/TR/sparql11-query/#extensionFunctions. Could be very useful to pass custom functions that SPARQL doesn't support.

jitsedesmet:


lib/Transformation.ts
```diff
@@ -156,25 +163,68 @@ function transformOperator(expr: Alg.OperatorExpression)
...
   if (!C.NamedOperators.contains(funcName as C.NamedOperator)) {
-    throw new Err.UnknownNamedOperator(expr.name.value);
+    if (!simpleAppl && !asyncAppl && config.syncExtensionFunctionCB) {
+      const func = config.syncExtensionFunctionCB(expr.name);
+      if (func) {
+        simpleAppl = wrapSyncFunction(func, expr.name.value);
+        return new E.SyncExtension(expr.name, args, simpleAppl);
+      }
+    }
+    if (!simpleAppl && !asyncAppl && config.asyncExtensionFunctionCB) {
```
rubensworks, line 212: We probably want to error early on if we receive both `syncExtensionFunctionCB` and `asyncExtensionFunctionCB`.

lib/Transformation.ts
```diff
@@ -156,25 +163,68 @@ function transformOperator(expr: Alg.OperatorExpression)
...
+    }catch (e) {
+      throw new ExtensionFunctionError(name, e);
+    }
+    return transformRDFTermUnsafe(res);
+  });
+}
+
 // TODO: Support passing functions to override default behaviour;
-export function transformNamed(expr: Alg.NamedExpression): E.NamedExpression {
+export function transformNamed(expr: Alg.NamedExpression, config: TransformAlgebraConfig)
```
rubensworks, line 198: First, instead of passing the new type `TransformAlgebraConfig`, I would suggest passing `AsyncEvaluatorConfig | SyncEvaluatorConfig`.

Second, this functions becomes a bit too complex IMO.
This now assumes that the config can contain both a async and async callback.
I would suggest to keep these two cases distinct (and error early on if a value for both is given).

If you keep them separate, you can simplify this function quite a bit.
Furthermore, you can then make the return type of this function conditional: if `AsyncEvaluatorConfig` is given, return `E.AsyncExtensionExpression`; if `SyncEvaluatorConfig` is given, return `E.SyncExtensionExpression`.

lib/Transformation.ts
```diff
@@ -1,20 +1,27 @@
...
-import { TypeURL as DT } from './util/Consts';
+import {namedFunctions, regularFunctions, specialFunctions,} from './functions';
+import {AsyncExtensionFunction, AsyncExtensionFunctionCB} from './evaluators/AsyncEvaluator';
+import {SyncExtensionFunction, SyncExtensionFunctionCB} from './evaluators/SyncEvaluator';
+import {ExtensionFunctionError} from './util/Errors';
 
-export function transformAlgebra(expr: Alg.Expression): E.Expression {
+export type ExtensionFunctionCB = AsyncExtensionFunctionCB | SyncExtensionFunctionCB;
+
+export interface TransformAlgebraConfig {
```
rubensworks, line 19: No need for this IMO (see comments below)

lib/evaluators/AsyncEvaluator.ts
```diff
@@ -1,23 +1,30 @@
...
 
-import { AsyncRecursiveEvaluator } from './RecursiveExpressionEvaluator';
+import {AsyncRecursiveEvaluator} from './RecursiveExpressionEvaluator';
+import {SyncExtensionFunctionCB} from './SyncEvaluator';
 
 type Expression = E.Expression;
 type Term = E.TermExpression;
 
+export type AsyncExtensionFunction = (args: RDF.Term[]) => Promise<RDF.Term>;
+export type AsyncExtensionFunctionCB = (functionNamedNode: NamedNode) => AsyncExtensionFunction;
```
rubensworks, line 17:
```suggestion
export type AsyncExtensionFunctionCallback = (functionNamedNode: NamedNode) => AsyncExtensionFunction;
```

lib/evaluators/AsyncEvaluator.ts
```diff
@@ -1,23 +1,30 @@
...
+
 export interface AsyncEvaluatorConfig {
   now?: Date;
   baseIRI?: string;
 
   exists?: (expression: Alg.ExistenceExpression, mapping: Bindings) => Promise<boolean>;
   aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;
   bnode?: (input?: string) => Promise<RDF.BlankNode>;
+  asyncExtensionFunctionCB?: AsyncExtensionFunctionCB;
```
rubensworks, line 26:
```suggestion
  extensionFunctionCallback?: AsyncExtensionFunctionCB;
```
(same name in `SyncEvaluatorConfig`)

rubensworks (changes requested):
Great start!

My main comment is regarding the distinction between sync and async, which can probably be simplified quite a bit.

lib/Transformation.ts
jitsedesmet: You're right, however, it has been mentioned in #16 that we might also want a custom implementation for functions. If we want to support this in the future isn't it good to have the argument object in place?

rubensworks: > we might also want a custom implementation for functions
> If we want to support this in the future isn't it good to have the argument object in place?

We can still enable this when having `AsyncEvaluatorConfig | SyncEvaluatorConfig` as argument, no?

jitsedesmet: No, it's probably cleaner to add a third argument than.

rubensworks: I see only two, what third one do you have in mind?

lib/Transformation.ts
jitsedesmet: I would like to keep `type ExtensionFunctionCallback = AsyncExtensionFunctionCallback | SyncExtensionFunctionCallback;` because the right part is half a line long and is quite present within the file. I think it serves it's purpose?

rubensworks: Sure, we can revisit in the next iteration.

lib/Transformation.ts
jitsedesmet: I meant when adding the custom function support.

lib/Transformation.ts
```diff
@@ -141,11 +143,12 @@ export function transformLiteral(lit: RDF.Literal): E.Literal<any> {
   }
 }
 
-function transformOperator(expr: Alg.OperatorExpression)
+function transformOperator<SyncType extends boolean>(expr: Alg.OperatorExpression,
```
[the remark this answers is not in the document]
wschella: (which I think you want to avoid anyway)

lib/expressions/Expressions.ts
```diff
@@ -11,6 +11,8 @@ export enum ExpressionType {
   SpecialOperator = 'specialOperator',
   Term = 'term',
   Variable = 'variable',
+  AsyncExtension = 'asyncExpression',
```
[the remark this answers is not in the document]
jitsedesmet: You're right. I don't know what I did here. The names should be improved.

lib/evaluators/AsyncEvaluator.ts
```diff
@@ -1,23 +1,27 @@
...
+import {Bindings, ExpressionEvaluator} from '../Types';
 
-import { AsyncRecursiveEvaluator } from './RecursiveExpressionEvaluator';
+import {AsyncRecursiveEvaluator} from './RecursiveExpressionEvaluator';
 
 type Expression = E.Expression;
 type Term = E.TermExpression;
 
+export type AsyncExtensionFunction = (args: RDF.Term[]) => Promise<RDF.Term>;
+export type AsyncExtensionFunctionCallback = (functionNamedNode: RDF.NamedNode) => AsyncExtensionFunction | null;
```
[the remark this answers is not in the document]
jitsedesmet: I think you're right in saying that callback might be vague and suggest unexpected behavior. I do like the name ´AsyncExtensionFunctionCreator´.

lib/Transformation.ts
```diff
@@ -1,31 +1,33 @@
...
-  specialFunctions,
-} from './functions';
-import { TypeURL as DT } from './util/Consts';
+import {namedFunctions, regularFunctions, specialFunctions,} from './functions';
+import {AsyncExtensionFunction, AsyncExtensionFunctionCallback} from './evaluators/AsyncEvaluator';
+import {SyncExtensionFunction, SyncExtensionFunctionCallback} from './evaluators/SyncEvaluator';
 
-export function transformAlgebra(expr: Alg.Expression): E.Expression {
-  if (!expr) { throw new Err.InvalidExpression(expr); }
+type ExtensionCallbackType<T extends boolean> = T extends true ? SyncExtensionFunctionCallback : AsyncExtensionFunctionCallback;
```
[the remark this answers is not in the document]
jitsedesmet: I don't mind changing it. I know @rubensworks had a vision for how this should work. @rubensworks what do you think of the proposed signatures?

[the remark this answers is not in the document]
rubensworks: Sure, I'm also fine with something like `ExtensionCallback = {type: 'sync', callback: SyncCallback} | {type: 'async', callback: AsyncCallback}`. No strong opinion on this myself :-)

lib/evaluators/AsyncEvaluator.ts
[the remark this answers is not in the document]
rubensworks: +1 on ´AsyncExtensionFunctionCreator´

lib/Transformation.ts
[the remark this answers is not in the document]
jitsedesmet: I think instead of calling it `ExtensionCallback` we might want t call it `UserFunctionCreatorConfig` if we start using `AsyncExtensionFunctionCreator´ this might make sense and we can keep this name when adding custom functions.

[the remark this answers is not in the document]
rubensworks: Sure, but perhaps just `FunctionCreatorConfig` then.

lib/Transformation.ts
```diff
@@ -190,20 +189,19 @@ function wrapAsyncFunction(f: AsyncExtensionFunction, name: string): AsyncExtens
...
 }
 // TODO: Support passing functions to override default behaviour;
-function transformNamed<SyncType extends  boolean>(expr: Alg.NamedExpression,
-                                                   extensionCallback: ExtensionCallbackType<SyncType>, sync: SyncType)
+function transformNamed(expr: Alg.NamedExpression, creatorConfig: FunctionCreatorConfig)
   : E.NamedExpression | E.AsyncExtensionExpression | E.SyncExtensionExpression  {
   const funcName = expr.name.value;
-  const args = expr.args.map((a) => transformAlgebra(a, extensionCallback, sync));
+  const args = expr.args.map((a) => transformAlgebra(a, creatorConfig));
   if (!C.NamedOperators.contains(funcName as C.NamedOperator)) {
```
wschella, line 196: Now we have this cleaned up a bit, we change the order of the paths and add some documentation.

Notice how the base case (a known named function) is now relegated to the end, with a lot of branching in between (even with negated expressions). I suggest moving this to the beginning, i.e. if it exists, great, return early. Then we remove a level of nesting for function extensions, which can come afterwards. Also add a little documentation to these branches.

wschella (changes requested):
- [ ] There should also be an example in the README on how to configure these extensions. Including a warning that if the name equals an existing name, nothing will happen.
- [ ] If issue #16 is closed, there should be made a new one for overriding behavior of existing functions.

wschella (approved):

rubensworks:
(I'll do another final review round once the proper hooks are in place in Comunica)

README.md
```diff
@@ -105,6 +106,20 @@ aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;
 
 You can probably ignore this.
 
+### Extension functions
+
+Extension functions can be added by providing the `extensionFunctionCreator` in the config.
+Example
+```ts
+config.extensionFunctionCreator = (functionName: RDF.NamedNode) => {
+   if (functionNamedNode.value === 'https://example.org/functions#equal') {
```
rubensworks, line 115: This may not be the most interesting example.
Could you add one for uppercasing a string instead?
(would then also require checking if the term is a literal)

README.md
```diff
@@ -105,6 +106,20 @@ aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;
...
+      return async (args: RDF.Term[]) => {
+         return literal(String(args[0].equals(args[1])), 'http://www.w3.org/2001/XMLSchema#boolean');       
+      }
+   }
+}
+```
+
 ### Custom functions
 
 **NOTE: Not yet implemented. On the roadmap.**
```
rubensworks, line 125: This can now go? :-)

lib/Transformation.ts
```diff
@@ -156,25 +158,61 @@ function transformOperator(expr: Alg.OperatorExpression)
...
-  if (!C.NamedOperators.contains(funcName as C.NamedOperator)) {
-    throw new Err.UnknownNamedOperator(expr.name.value);
+  const args = expr.args.map((a) => transformAlgebra(a, creatorConfig));
+  if (C.NamedOperators.contains(funcName as C.NamedOperator)) {
+    // return a basic named expression
+    const op = expr.name.value as any as C.NamedOperator;
+    const func = namedFunctions.get(op);
+    return new E.Named(expr.name, args, func.apply);
+  } else if (creatorConfig.type === 'sync') {
+    // Expression might be extension function, check this for the sync or async case
```
rubensworks, line 202: Comment incorrect?
We're in the sync case?
(I'd add a similar comment in the async case)

lib/evaluators/AsyncEvaluator.ts
```diff
@@ -1,23 +1,27 @@
...
+import {Bindings, ExpressionEvaluator} from '../Types';
 
-import { AsyncRecursiveEvaluator } from './RecursiveExpressionEvaluator';
+import {AsyncRecursiveEvaluator} from './RecursiveExpressionEvaluator';
 
 type Expression = E.Expression;
 type Term = E.TermExpression;
 
+export type AsyncExtensionFunction = (args: RDF.Term[]) => Promise<RDF.Term>;
+export type AsyncExtensionFunctionCreator = (functionNamedNode: RDF.NamedNode) => AsyncExtensionFunction | null;
```
rubensworks, line 15: Just wondering, would the following be (too) complicated to support?
```suggestion
export type AsyncExtensionFunctionCreator = (functionNamedNode: RDF.NamedNode) => Promise<AsyncExtensionFunction | null>;
```

rubensworks: Also, can we use `undefined` everywhere instead of `null`, as explained in the Comunica PR?

test/functions/extension.test.ts
```diff
@@ -0,0 +1,57 @@
...
+import * as RDF from 'rdf-js';
+import {NamedNode} from 'rdf-js';
+
+function mapToTerm(map: StringMap): TermMap {
+  return Object
+    .keys(map)
+    .reduce((p, k) => ({ ...p, [k]: stringToTermPrefix(map[k]) }), {});
+}
+
+describe('extension function: term-equal', () => {
```
rubensworks, line 12: Can we also have a test case that checks a custom uppercase function?

rubensworks (changes requested):
Just some minor comments

README.md
jitsedesmet: If you `custom function`, I don't think so, as Wout said, we should make a distinction between custom function and extension function. What we did is the extension function. I think we should leave it there because we are still planning on implementing custom functions right?

lib/Transformation.ts
jitsedesmet: I thought that might've been overkill, I'll add another.

README.md
rubensworks: Well, at the moment, most readers of the README will not know the difference between the two, and this will cause confusion.

I think the "Custom functions" section was meant for the extension functions we have here.
The custom functions you refer to would be more like function _overrides_, which can indeed be added in the future, but doesn't seem that critical do add in the README as something that is coming soon.

lib/evaluators/AsyncEvaluator.ts
jitsedesmet: I don't think it would be, not sure. I'm not sure why you would want this. Anyway I see to ways in handeling this.
1. We make the constructor async and everything else stays the same.
2. In transform named we call the creator function. This gives us a promise which we await in the `evalAsyncExtension` function. Problem here is that when the promise revolves to be null/ undefined, we would need to throw the `UnknownNamedOperator` error or something, this behavior is completely different from what we have now. Right now we expect that we know whether we can execute an expression on the creation of the evaluator. This brings us to the biggest problem. Libraries like comunica use the creator as a test to check if a query is executable. This assertion would not be valid anymore.

rubensworks: Ok, let's leave it for future work then.

README.md
jitsedesmet: You'e right, I'll remove it from the README.

lib/evaluators/AsyncEvaluator.ts
jitsedesmet: looking at the first suggestion I made. It's impossible to make a constructor async. We would need something like `async readyAsyncEvaluator()` to handle this than. What are your thoughts on this? When we do this we should not forget to edit the README.

rubensworks: Let's just leave it out for now then, It's not that important to have for now :-)

jitsedesmet: I just thought of another solution. We could make a hybrid of 1 and 2. We can do everything 2 would do and just provide something like `async readyEvaluation` which awaits the promise (NOTE that calling this function would be optional). Comunica might than be able to call this function in it's test. It shouldn't be to hard to implement this.

rubensworks: Seems too error-prone, would have to be something mandatory.
But I wouldn't worry about it too much now, we can revisit it via a separate PR if needed.

jitsedesmet: Alright! I'll resolve the conversation than.

test/functions/extension.test.ts
jitsedesmet: This would be pretty hard to implement as of right now but we should probably test this when we look at the tests.

README.md
jitsedesmet: Since this would be hard to test as of right now maybe we should leave this example as is? We can update it when we can also test this example?

rubensworks: Sounds good!

README.md
```diff
@@ -41,6 +41,7 @@ interface AsyncEvaluatorConfig {
   exists?: (expression: Alg.ExistenceExpression, mapping: Bindings) => Promise<boolean>;
   aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;
   bnode?: (input?: string) => Promise<RDF.BlankNode>;
+  extensionFunctionCreator?: (functionNamedNode: RDF.NamedNode) => (args: RDF.Term[]) => Promise<RDF.Term> | null;;
```
rubensworks, line 44:
```suggestion
  extensionFunctionCreator?: (functionNamedNode: RDF.NamedNode) => (args: RDF.Term[]) => Promise<RDF.Term> | undefined;
```

PR #88, opened 2021-07-29
Feature/async aggregate evaluator (closes #86)

### LINKED ISSUE #86: AgragateEvaluator uses SyncEvaluator
@wschella I was wondering why `AgragateEvaluator` uses `SyncEvaluatorConfig`.
I was trying to make sure the custom-function implementation for #16 was usable in a bigger project like comunica. Within this project almost everything is async, so it wouldn't make any sense to ask a developer for sync extension functions. However, the project uses sparqlee's `AggragateEvaloator` which is build apon the `SyncEvaluator`. Maybe we should make an `AsyncAggragateEvaloator` for completeness?

jitsedesmet:
Iimplementation of an async version of the existing `AggragateEvaluator` as explained in #86 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,11 @@ aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;
 
 You can probably ignore this.
 
+We also provide an `AsyncAggregateEvaluator` to that works the same way `AggregateEvaluator` does.
+The signature of only the `put` methode changes to be async. It is up to you to handle this correctly.
+You are for example expected to await all puts before you ask for `result`.
+You should also note the order of calling and awaiting put while using the `GroupConcat` aggregator.
+
 ### Extension functions
 
 Extension functions can be added by providing the `extensionFunctionCreator` in the config.
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;
 You can probably ignore this.
 
 We also provide an `AsyncAggregateEvaluator` to that works the same way `AggregateEvaluator` does.
-The signature of only the `put` methode changes to be async. It is up to you to handle this correctly.
+The signature of only the `put` method changes to be async. It is up to you to handle this correctly.
 You are for example expected to await all puts before you ask for `result`.
 You should also note the order of calling and awaiting put while using the `GroupConcat` aggregator.
 
--- a/lib/evaluators/AggregateEvaluator.ts
+++ b/lib/evaluators/AggregateEvaluator.ts
@@ -21,13 +21,13 @@ import {AsyncEvaluator, AsyncEvaluatorConfig} from './AsyncEvaluator';
 
 const DF = new DataFactory();
 
-abstract class BaseAggregateEvaluator {
+class BaseAggregateSupporter {
   protected expression: Algebra.AggregateExpression;
   protected aggregator: BaseAggregator<any>;
   protected throwError = false;
   protected state: any;
 
-  protected constructor(expr: Algebra.AggregateExpression, throwError?: boolean) {
+  constructor(expr: Algebra.AggregateExpression, throwError?: boolean) {
     this.expression = expr;
     this.aggregator = new aggregators[expr.aggregator as SetFunction](expr);
     this.throwError = throwError;
@@ -49,6 +49,10 @@ abstract class BaseAggregateEvaluator {
     return val;
   }
 
+  result(): RDF.Term {
+    return (this.aggregator.constructor as AggregatorClass).emptyValue();
+  }
+
   /**
    * Put a binding from the result stream in the aggregate state.
    *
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -106,6 +106,11 @@ aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;
 
 You can probably ignore this.
 
+We also provide an `AsyncAggregateEvaluator` to that works the same way `AggregateEvaluator` does.
+The signature of only the `put` methode changes to be async. It is up to you to handle this correctly.
```
rubensworks, line 110:
```suggestion
The signature of only the `put` method changes to be async. It is up to you to handle this correctly.
```

rubensworks:
Looks good at first glance!
Could you check if this works as intended in Comunica (perhaps with a good test case)? I can have a more detailed look at this afterwards.

wschella (changes requested):
I think I implemented the init/put/__put state weirdly, or I don't understand it anymore. In any case something should be fixed, and this PR is an opportune moment (sorry for putting the burden on you ;) )

Now, I wonder why 
```
this.put = this.__put;
this.result = this.__result;
```
is in `put(..)`, and not in the `try` branch of `init(..)` (for the Sync Aggregator). I think this would clean things up. Simmilarly, I think the Async version should also have an `init(..)` function, just to be symmetrical. 

Then all the initial `put` has to do is call `init`, which would IMO be more clear to the developer.

Also, do we expect all `puts` to be awaited in order, or does order not matter. Cause for `groupconcat` that's a big difference. Should be made clear in the docs.

jitsedesmet:
You're right, I removed the init because that didn't make sense for me either. You're right however that it would be cleaner if put just calls init and does nothing else. I thought the diffrence for `GroupConcat` was made clear in the README?

wschella:
> You're right, I removed the init because that didn't make sense for me either. You're right however that it would be cleaner if put just calls init and does nothing else. I thought the diffrence for `GroupConcat` was made clear in the README?

I'm sorry, I read over the README changes somehow.

jitsedesmet:
I just pushed a commit in which I also added a function that allows users to evaluate a term and put the term in the aggregator themself with putTerm. I did this because evaluation, especially with the recently added extension functions in async mode, could take some time. This allows a user to control this evaluation when he/she would feel the need for it.

wschella (approved):

jitsedesmet:
@wschella I just bumped into some weird behavior which made me wonder, why doesn't safeThrow set a boolean, 'errorOccured' which the evaluatpr checks?

lib/evaluators/AggregateEvaluator.ts
```diff
@@ -94,29 +87,154 @@ export class AggregateEvaluator {
...
+    } else {
+      this.putTerm = () => { return; };
+      this.result = () => undefined;
+    }
+  }
+}
+
+abstract class BaseAggregateEvaluator extends BaseAggregateSupporter {
+
+  abstract evalualte(bindings: Bindings): RDF.Term | Promise<RDF.Term>;
```
wschella, line 118: Typo

lib/evaluators/AggregateEvaluator.ts
```diff
@@ -94,29 +87,154 @@ export class AggregateEvaluator {
...
+  /**
+   * Put a binding from the result stream in the aggregate state.
+   *
+   * If any binding evaluation errors, the corresponding aggregate variable should be unbound.
+   * If this happens, calling @see result() will return @constant undefined
+   *
+   * @param bindings the bindings to pass to the expression
+   */
+  abstract put(bindings: Bindings): void | Promise<void>;
+  /**
```
wschella, line 129: Needs a newline

wschella:
> @wschella I just bumped into some weird behavior which made me wonder, why doesn't `safeThrow` set a boolean, 'errorOccured' which the evaluator checks?

I don't quite understand the question. The reason it's implemented like this is likely some spec thing I should have documented. I think in spec defined behavior when an error occurs they aren't supposed to be thrown and the result is simply undefined. That's what implemented + a setting to throw errors if you explicitly do want them. 

Whether that's wishful is one thing, but note you can have many many errors, and silent errors for some parts of the stream are true footguns, so this isn't too bad I think.

jitsedesmet:
Yes, I now realise my question was actually not really valid for the code you implemented. It was a way async silent errors are to be handled here it is necessary to have a boolean. Don't worry about it ;) (well I hope)

README.md
```diff
@@ -106,6 +106,11 @@ aggregate?: (expression: Alg.AggregateExpression) => Promise<RDF.Term>;
 
 You can probably ignore this.
 
+We also provide an `AsyncAggregateEvaluator` to that works the same way `AggregateEvaluator` does.
+The signature of only the `put` method changes to be async. It is up to you to handle this correctly.
+You are for example expected to await all puts before you ask for `result`.
+You should also note the order of calling and awaiting put while using the `GroupConcat` aggregator.
```
rubensworks, line 112: `GroupConcat` doesn't exist in Sparqlee, to this comment does not make much sense here.

rubensworks (changes requested):
Nice work! Will merge together with the comunica PR.

README.md
wschella: It exists in Sparqlee: https://github.com/comunica/sparqlee/blob/6c609b165e48bb190042b1fdbaf019e2c78d59f5/lib/evaluators/AggregateEvaluator.ts#L266

rubensworks: Nvm my comment then, I thought it was part of Comunica.
See, that's how well I know how the expressions work 😅

rubensworks (approved):

PR #89, opened 2021-08-02
Write better example use documentation in README

wschella:


rubensworks (approved):
Much clearer indeed, thanks!

Issue #40, opened 2019-05-20
Fix the spec/functions/replace03 test


wschella:
Due to the annoying syntax of the replace03 test, our test parsing fails, so it's disabled for now.

Issue #34, opened 2019-05-20
Clean up tests


wschella:
- [x] Testing between operators is inconsistent
- [ ] Util folder needs documentation and cleanup
- [ ] Util.test needs to be renamed

jitsedesmet:
I was looking at this issue and I wonder whether you could provide some more information on the tasks you mentioned.
I was also looking at the tests and noticed there's quite some similarity between `testAll` in combination with `testAllErrors` and the classes in the `Truthtable` file. I wonder if you could clarify the difference between them? If they are historically developed and provide (almost) the same service, I would want to remove them. Less is more after all.

wschella:
I think it's a bit of both, testAll+testAllErrors is a bit more straightforward (in implementation and use) then the TruthThable, it doesn't require the aliasing maps, and doesn't hide the operator. The cases in testAll(Errors) are simply correct expressions, while the Truthtable is more like a parametrized expression. It think it's sensible to have both, but this difference is not very clear from naming. documentation, and API.

If you can abstract this decently, I'm definitely not opposed to it, e.g. with naming `testAll`, `testAllErrors`, and `testAllParametrized` would already make more sense. If you can then make the API a bit more similar and reuse some logic. I bet there's inspiration to be gained from test framework conventions.

In regards to the points in the issues:
- []  Testing between operators is inconsistent. I have no clue anymore.
- [] Util folder needs documentation and cleanup. As you can see, the test organization is a bit of a mess. I managed to get a nice structure for _writing_ tests easily, and then I had enough, but behind the hood it needs structure & clarity. I think most files need at least some documentation on how they are used. The organization can be cleaned up. You know, make this something you can confidently pass to the next person. 
- [] Util.test needs to be renamed. This issue is broader, the name "util.test" exists only to differentiate from the "util" folder, which is utilities for testing. I think following structure might make more sense:

```
- lib/    # Testing structure matching the lib organization
  - evaluators/
  - functions/
  - ... 
- spec/   # The spec tests
- util/   # Test utilities
```

where `misc` can be at the root, or in `lib/`

jitsedesmet:
Thank you for the clarification! I will consider `Testing between operators is inconsistent` as done unless I were to suddenly understand what you meant by it.

rubensworks:
I would suggest keeping everything dist-related in `lib/`, and everything test-related in `test/`
So perhaps something more like this:

```
- lib/
  - evaluators/
  - functions/
  - util/
  - ... 
- test/
[...]

PR #97, opened 2021-08-05
Clean/generalize test into one backbone

jitsedesmet:
This pull request aims to clean the test suit and provides a little more explanation on how tests work and how you can write new ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,29 +343,55 @@ Running tests will generate a `test-report.html` in the root dir.
 
 The testing environment is set up to do a lot of tests with little code.
 The files responsible for fluent behaviour reside in `test/util`.  
-Most test can be run by running the `runTestTable` method in `test/util/utils.ts`.
-This method expects a TruthTable. Multiple test are run over a TruthTable (one for every line).
-A truth table may contain aliases if the aliases are also provided
+Most tests can be run by running the `runTestTable` method in `test/util/utils.ts`.
+This method expects a TestTable. Multiple test are run over a TestTable (one for every line).
+A TestTable may contain aliases if the aliases are also provided
 (Some handy aliases reside in `test/util/Aliases.ts`).
 This means that when testing something like `"3"^^xsd:integer equals "3"^^xsd:integer` is `"true"^^xsd:boolean`.
-We would write a small table and test it like this:
+We would write a small table (for this example some more tests are added) and test it like this:
 ```ts
 import { bool, merge, numeric } from './util/Aliases';
 import { Notation } from './util/TruthTable';
 import { runTestTable } from './util/utils';
 runTestTable({
+   testTable: `
+       3i 3i = true
+       3i -5i = false
+       -0f 0f = true
+       NaN  NaN = false
+   `,
    arity: 2,
    operation: '=',
    aliases: merge(numeric, bool),
    notation: Notation.Infix,
-   testTable: `
-    3i 3i = true
-   `,
 });
 ```
-More options can be provided and are explained with the type definition of the argument of `runTestTable`. 
+More options can be provided and are explained with the type definition of the argument of `runTestTable`.
 
-In case the tables are too restrictive for your test and you need an evaluation.
+We can also provide an `errorTable` to the `runTestTable` method.
+This is used when we want to test if calling certain functions on certain arguments throws the error we want.
+An example is testing whether `Unknown named operator` error is thrown when
+we don't provide the implementation for an extension function.
+```ts
+import { bool, merge, numeric } from './util/Aliases';
+import { Notation } from './util/TruthTable';
+import { runTestTable } from './util/utils';
+runTestTable({
+   errorTable: `
+       3i 3i = 'Unknown named operator'
+       3i -5i = 'Unknown named operator'
+       -0f 0f = 'Unknown named operator'
+       NaN  NaN = 'Unknown named operator'
+   `,
+   arity: 2,
+   operation: '<https://example.org/functions#equal>',
+   aliases: merge(numeric, bool),
+   notation: Notation.Infix,
+});
+```
+When you don't care what the error is, you can just test for `''`.
+
+In case the tables are too restrictive for your test, and you need an evaluation.
 You should still use the `generalEvaluate` function from `test/util/generalEvaluation.ts`.
 This function will automatically run both async and sync when possible.
 This increases your tests' coverage.
--- a/README.md
+++ b/README.md
@@ -340,7 +340,32 @@ We also handle type promotion by explicitly coding for it. This is done in the a
 ### Testing
 
 Running tests will generate a `test-report.html` in the root dir.
-**TODO** Explain test organizatian and expression tables
 
-Evaluating within a test should be done by calling `generalEvaluate`.
-This will test both sync and async implementation if possible. 
+The testing environment is set up to do a lot of tests with little code.
+The files responsible for fluent behaviour reside in `test/util`.  
+Most test can be run by running the `runTestTable` method in `test/util/utils.ts`.
+This method expects a TruthTable. Multiple test are run over a TruthTable (one for every line).
+A truth table may contain aliases if the aliases are also provided
+(Some handy aliases reside in `test/util/Aliases.ts`).
+This means that when testing something like `"3"^^xsd:integer equals "3"^^xsd:integer` is `"true"^^xsd:boolean`.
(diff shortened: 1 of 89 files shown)


jitsedesmet:
I could just re-enable the test disabled in #40 so this PR should close that issue.

README.md
```diff
@@ -340,7 +340,32 @@ We also handle type promotion by explicitly coding for it. This is done in the a
...
+import { bool, merge, numeric } from './util/Aliases';
+import { Notation } from './util/TruthTable';
+import { runTestTable } from './util/utils';
+runTestTable({
+   arity: 2,
+   operation: '=',
+   aliases: merge(numeric, bool),
+   notation: Notation.Infix,
+   testTable: `
+    3i 3i = true
```
wschella, line 362: Nitpick. Let's put the testTable argument first, and add some more examples to highlight the use of it. We want to convince future developers of this approach!

test/util/Aliases.ts
```diff
@@ -1,70 +1,72 @@
 import type * as RDF from 'rdf-js';
 import { stringToTerm } from 'rdf-string';
 
-import type { GeneralEvaluationConfig } from './generalEvaluation';
-import type { IEvaluationConfig, Notation } from './TruthTable';
+/**
+ * Maps short strings to longer RDF term-literals for easy use in making evaluation tables.
+ * Ex: { 'true': '"true"^^xsd:boolean' }
+  */
```
wschella, line 7: Formatting. Also, some off these functions could use a line of documentation or an example (especially the latter).

README.md
```diff
@@ -340,7 +340,32 @@ We also handle type promotion by explicitly coding for it. This is done in the a
...
 
 Running tests will generate a `test-report.html` in the root dir.
-**TODO** Explain test organizatian and expression tables
 
-Evaluating within a test should be done by calling `generalEvaluate`.
-This will test both sync and async implementation if possible. 
+The testing environment is set up to do a lot of tests with little code.
+The files responsible for fluent behaviour reside in `test/util`.  
+Most test can be run by running the `runTestTable` method in `test/util/utils.ts`.
+This method expects a TruthTable. Multiple test are run over a TruthTable (one for every line).
```
wschella, line 347: TruthTable is not a class anymore (the file is still called that way). Anyway lets rename it to TestTable or smth?

test/util/utils.ts
```diff
@@ -1,88 +1,70 @@
...
-    it(`${body} should evaluate to ${_result}`, async() => {
-      const evaluated = await generalEvaluate({
-        expression: template(body), expectEquality: true, generalEvaluationConfig: config,
-      });
-      expect(termToString(evaluated.asyncResult)).toEqual(termToString(result));
-    });
-  });
+export interface ITestArgumentBase {
+  /**
+   * Operation / function that needs to be called on the arguments provided in the truthTable.
```
wschella, line 8: lowercase `truth`

test/util/utils.ts
```diff
@@ -1,88 +1,70 @@
...
-    const result = stringToTerm(replacePrefix(_result));
-    // Console.log(`${expr}\n${equals}\n${body}\n${_result}\n${result}`);
-    it(`${body} should evaluate to ${_result}`, async() => {
-      const evaluated = await generalEvaluate({
-        expression: template(body), expectEquality: true, generalEvaluationConfig: config,
-      });
-      expect(termToString(evaluated.asyncResult)).toEqual(termToString(result));
-    });
-  });
+export interface ITestArgumentBase {
```
wschella, line 6: I don't really like the name. What about ITestTableConfigBase?

wschella (changes requested):
I wonder if we can replace the config with a template.
So instead of 
```
notation: Notation.Infix,
operation: '+',
arity: 2,
```
we can do something like `template: "$TERM + $TERM"`. Anyway, only do this if it's easy and you think it's interesting. Cause it's not necessary, just fun maybe?

This pull request is great! Much needed cleanup, you've done a good job.

jitsedesmet:
I guess a template would be pretty cool. Personally I like the current syntax because it provides a nice description on the function. A template is a bit harder for `arity: 'vary'` but that one is a bit of a weird one anyway :/ . Anyway, I personalty like the description syntax we have now a bit more.

README.md
jitsedesmet: What expamples do you mean? The same test but new lines to show how multiple tests in the table are created?

README.md
jitsedesmet: Seems like a decent name, sure! I wanted to change the file structure anyway so maybe I should just do that in this PR. (Changes we talked about in #34)

test/util/utils.ts
jitsedesmet: Sure!

README.md
wschella: Yes exactly! It's easier to abstract from multiple demonstrations :)

wschella:
> I guess a template would be pretty cool. Personally I like the current syntax because it provides a nice description on the function. A template is a bit harder for `arity: 'vary'` but that one is a bit of a weird one anyway :/ . Anyway, I personalty like the description syntax we have now a bit more.

It's true, it showcases clearly which properties of the function are relevant for the logic behind it. Let's keep it without templates.

test/util/TruthTable.ts
```diff
@@ -1,159 +1,188 @@
...
+  Suffix,
   Function,
 }
 
-// Given the definition for an evaluation table, test all it's test cases.
-export function testTable(definition: EvaluationTable): void {
-  const tTable = (definition.arity === 2) ?
-    new BinaryTable(definition, BinaryTableParser) :
-    new UnaryTable(definition, UnaryTableParser);
+type Row = [string, string, string] | [string, string] | string[];
```
rubensworks, line 13: Any difference with this?
```suggestion
type Row = string[];
```

test/util/TruthTable.ts
```diff
@@ -1,159 +1,188 @@
...
+
+/**
+ * A series of tests in string format.
+ * A difference is made between erroring and non-erroring tests-cases because
+ * the testing framework requires it, and it allows for some reduction in
+ * boilerplate when errorhandling is similar between operators.
+ * Some mappings are passed to translate between shorthand aliases and their
+ * full SPARQL representations.
+ *
+ * It's format is
```
rubensworks, line 23:
```suggestion
 * Its format is
```

test/util/utils.ts
```diff
@@ -1,88 +1,70 @@
...
+  /**
+   * Operation / function that needs to be called on the arguments provided in the truthTable.
+   */
+  operation: string;
+  /**
+   * How many arguments does the operation take. The vary option means you don't know. This can only be provided
+   * when the notation is Notation.function.
```
rubensworks, line 13:
```suggestion
   * when the notation is Notation.Function.
```

rubensworks (changes requested):
Nice cleanup!
No major comments from my end.

test/util/TruthTable.ts
jitsedesmet: Not really. I kept it because I thought it would show the difference between the 3 more but it doesn't really matter.

wschella (approved):

test/util/TruthTable.ts
wschella: I do prefer it like this as well, highlights the three cases.

rubensworks: AFAIK it won't make any difference for autocompletion within IDEs, but no strong opinion eitherway.

test/util/Aliases.ts
jitsedesmet: I don't think I understand what you mean.

test/util/TruthTable.ts
wschella: It doesn't indeed. It's committed like this now, let's keep it as such.

test/util/Aliases.ts
wschella: Ah it's outdated now, but there was a space missing.

For the rest, these utility functions (below, I was lazy and bundled my feedback) really benefit from having a comment that provides an example of input/output. Since this is becoming a general test cleanup PR anyway, I thought it would make sense (but it's not required, just nice to have, so as you wish).

jitsedesmet: Well I added some explanation in the last commit so should be fine now anyways :D

Issue #94, opened 2021-08-04
Enable `no-redeclare` rule


jitsedesmet:
While trying to fix #93 it became clear I needed to disable the `no-redeclare` rule because an implementation of `Set` is used in the file `lib/util/Consts.ts` I couldn't just use the default implementation of Set because a `union` function is used. This issue, tho probably small aims to enable the `no-redeclare` rule.

jitsedesmet:
Since it's also time to move away from immutable I will do this in this issue as well. This issue will thus also remove all References to the `Immutable` package.

jitsedesmet:
I stumbled into a problem moving away from Immutable. In `lib/functions/Core.ts` we use an something called `export type OverloadMap = Map<List<ArgumentType>, E.SimpleApplication>;`. The basic transition would be to write `export type OverloadMap = Map<ArgumentType[], E.SimpleApplication>;` instead. Problem with this is that array doesn't define the kind of equality we want here. I don't really understand this code very well yet. So I'm already asking, do you (in particular @wschella) know any good solutions for this.

jitsedesmet:
I also don't think @rubensworks will let me write my own equals within the Array prototype :p .

wschella:
This is the indeed some difficult code, as we basically define a DSL for function overloading in Typescript in type safe way. Lemme check! In the meanwhile, what's the reason with moving away from immutable (i'm not up to date with the ecosystem)?

Context: A couple of years ago I think Immutable was heavily used in comunica, and @rubensworks was definitely fan of using it in Sparqlee as well. I needed a List object that was a valid entry for a HashMap key (which a standard JS is not). I'll have a look at the code, I have some ideas.

jitsedesmet:
I don't really know the reason. Maybe @rubensworks isn't that good of a friend with them anymore. I love that you have some ideas because I am fresh out of them? Btw, what does DSL stand for? (sorry)

wschella:
So the general gist behind OverloadMap is that SPARQL functions are overloaded (the same function has different amount of arguments, or the arguments can be different types). OverloadMap keeps an implementation of this function for each overload. Overloads are uniquely defined by the list of arguments to the function and their types (which is runtime information). So when we know at runtime which concrete types are passed to our function, we can look up the corresponding definition in the overloadmap. 

[...]

PR #100, opened 2021-08-09
Issue/94 enable no redeclare

jitsedesmet:
This pull request aims to resolve most of #94. Except for the files `lib/functions/Helpers.ts` and `lib/functions/Core.ts`. These files cover the problem also talked about in #94 regarding the `OverloadMap`. Rewriting `OverloadMap` will be a pull request on itself.

### DIFF
--- a/lib/Types.ts
+++ b/lib/Types.ts
@@ -10,8 +10,9 @@ export type Bindings = Map<string, RDF.Term>;
  * @return {Bindings} The immutable bindings from the hash.
  * @constructor
  */
-export function BindingsFromHash(hash: Record<string, RDF.Term>): Bindings {
-  return Map(Object.entries(hash));
+// eslint-disable-next-line no-redeclare
+export function Bindings(hash: Record<string, RDF.Term>): Bindings {
+  return Map(hash);
 }
 
 export interface IExpressionEvaluator<ExpressionType, TermType> {
--- a/lib/Types.ts
+++ b/lib/Types.ts
@@ -1,3 +1,5 @@
+// eslint-disable-next-line no-redeclare
+import { Map } from 'immutable';
 import type * as RDF from 'rdf-js';
 
 export type Bindings = Map<string, RDF.Term>;
@@ -9,7 +11,7 @@ export type Bindings = Map<string, RDF.Term>;
  * @constructor
  */
 export function BindingsFromHash(hash: Record<string, RDF.Term>): Bindings {
-  return new Map(Object.entries(hash));
+  return Map(Object.entries(hash));
 }
 
 export interface IExpressionEvaluator<ExpressionType, TermType> {
--- a/lib/Types.ts
+++ b/lib/Types.ts
@@ -1,4 +1,3 @@
-import { Map } from 'immutable';
 import type * as RDF from 'rdf-js';
 
 export type Bindings = Map<string, RDF.Term>;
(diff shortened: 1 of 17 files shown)


lib/Types.ts
```diff
@@ -9,8 +8,8 @@ export type Bindings = Map<string, RDF.Term>;
  * @return {Bindings} The immutable bindings from the hash.
  * @constructor
  */
-export function Bindings(hash: Record<string, RDF.Term>): Bindings {
-  return Map(hash);
+export function BindingsFromHash(hash: Record<string, RDF.Term>): Bindings {
```
rubensworks, line 11: This (`Bindings`) is apparently literally taken over from Comunica.
Not sure how strict the relationship to Comunica is with respect to this, but can you double-check if everything still works after this change?

In any case, I see no need to rename this function.

lib/functions/NamedFunctions.ts
```diff
@@ -27,7 +25,7 @@ type Term = E.TermExpression;
 // https://www.w3.org/TR/xpath-functions/#casting-from-primitive-to-primitive
 // ----------------------------------------------------------------------------
 
-const toString = {
+const toStringNamed = {
```
rubensworks, line 28: Why did you rename these?

lib/functions/RegularFunctions.ts
```diff
@@ -171,50 +170,50 @@ const greaterThanEqual = {
 // https://www.w3.org/TR/sparql11-query/#func-rdfTerms
 // ----------------------------------------------------------------------------
 
-const isIRI = {
+const rdfTermIsIRI = {
```
rubensworks, line 173: Same as above, rename is not needed?

lib/functions/RegularFunctions.ts
```diff
@@ -793,4 +792,4 @@ export interface IDefinition {
   overloads: OverloadMap;
 }
 
-export const definitions = Map<C.RegularOperator, IDefinition>(_definitions);
+export const definitions = _definitions;
```
rubensworks, line 795: Just export the definitions above?

rubensworks (changes requested):

lib/Types.ts
jitsedesmet: Well, the linter doesn't like a function Bindings and a type Bindings to have the same name. it does seem like comunica has some problems with this PR tho. I should investigate that closer.

lib/functions/NamedFunctions.ts
jitsedesmet: I don't really know, where but according to the linter this function already exists in the global-scope.

lib/functions/RegularFunctions.ts
jitsedesmet: This one is indeed not needed, but, since there was also a `toString` function (now `rdfTermToString`) I renamed all the functions so they still have the same naming style.

lib/functions/RegularFunctions.ts
jitsedesmet: Yes, we could do that.

lib/Types.ts
jitsedesmet: Seems like `BindingsFromHash` Isn't used in comunica. `Bindings` however is (in the config object, within exists). So I just made that an immutable Map again.

rubensworks: Ok, good!

Could you restore the implementation as well to what it was before?

And I would recommend keeping the name `Bindings` for consistency with Comunica. It's ok to add a linter exception for this.

jitsedesmet: Okey!

lib/functions/RegularFunctions.ts
wschella: In this case I would not rename any, except `toString` and just document that it's an en exception. I would say use `termToString` (rdf is unneeded here), `toString_`, or even `str` which is what it's called in SPARQL.

wschella (changes requested):

rubensworks (approved):
Great work @jitsedesmet!

PR #101, opened 2021-08-10
Rewrite/overloadmap as tree

jitsedesmet:
This pull request aims to provide a first implementation for transforming the overloadmap to a tree structure. This removes it dependency on the 'immutable' package and should allow us to auto cast some concrete types.
In the first version of this PR I added 3 TODO's. These should not be merged into master but should be discussed here and resolved before merging.


lib/functions/Core.ts
```diff
@@ -12,21 +8,69 @@ type Term = E.TermExpression;
 // Overloaded Functions
 // ----------------------------------------------------------------------------
 
-// Maps argument types on their specific implementation.
-export type OverloadMap = Map<List<ArgumentType>, E.SimpleApplication>;
+type SearchStack = OverloadNode[];
+
+export class OverloadNode {
```
rubensworks, line 13: Let's move this to a separate file.

rubensworks: Also, since it is not trivial to understand what is going on in this class, I would suggest adding brief jsdoc to explain why it exists, and what it does.

lib/functions/Core.ts
```diff
@@ -45,6 +89,27 @@ export abstract class BaseFunction<Operator> {
...
 
+  /**
+   * Searches in a depth fist way for the best matching overload.
+   * @param args
+   * @private
+   */
+  private search(args: Term[]): E.SimpleApplication {
+    const searchStack: { node: OverloadNode; index: number }[] = [];
+    const startIndex = 0;
+    searchStack.push(...this.overloads.getSubTreeWithArg(args[startIndex]).map(node =>
```
rubensworks, line 100: Why is the copy needed?

lib/functions/Core.ts
```diff
@@ -45,6 +89,27 @@ export abstract class BaseFunction<Operator> {
...
+      ({ node, index: startIndex + 1 })));
+    while (searchStack.length > 0) {
+      const { index, node } = <{ node: OverloadNode; index: number }> searchStack.pop();
+      if (index === args.length) {
+        return node.getImplementation();
+      }
+      searchStack.push(...node.getSubTreeWithArg(args[index]).map(item =>
```
rubensworks, line 107: Same question, why copy?

lib/functions/Core.ts
```diff
@@ -45,6 +89,27 @@ export abstract class BaseFunction<Operator> {
 
   protected abstract handleInvalidTypes(args: Term[]): never;
 
+  /**
+   * Searches in a depth fist way for the best matching overload.
```
rubensworks, line 93: Typo

lib/functions/Core.ts
```diff
@@ -45,6 +89,27 @@ export abstract class BaseFunction<Operator> {
...
+   * @param args
+   * @private
+   */
+  private search(args: Term[]): E.SimpleApplication {
```
rubensworks, line 97: Does this needs checks in case `getImplementation` is `undefined`?

lib/functions/Core.ts
```diff
@@ -12,21 +8,69 @@ type Term = E.TermExpression;
...
+    }
+    this.subTrees[str]._addOverload(argumentTypes, func);
+  }
+
+  /**
+   * @param arg term to try and match to possible overloads of this node.
+   * @returns SearchStack a stack with top element the next node that should be asked for implementation or overload.
+   */
+  public getSubTreeWithArg(arg: Term): SearchStack {
+    const matching = Object.entries(this.subTrees).map(([ type, node ]) =>
```
rubensworks, line 49: This parts needs some cleaning and documentation (any-cast, magic constants, nested ternary, unneeded casts).

rubensworks (changes requested):
I've added some comments, but I don't fully understand what the code does; likely due to the minimal documentation.

lib/functions/Core.ts
jitsedesmet: No, this should also return `| undefined` my bad.

rubensworks: We should urgently enable strict compilation in tsconfig, to avoid issues like this.

lib/functions/Core.ts
jitsedesmet: The magic constants are just for now actually, they would go away with the implementation of autocasts. The casts are necessary I think because sort thinks the type of the array is `(number | OverladNode)[]` when it is in fact `[number, OverladNode]`. For the any cast, I have no idea, it was used like that in the `Typer` object and I just took it from there.

lib/functions/Core.ts
jitsedesmet: Alright!

lib/functions/Core.ts
jitsedesmet: We don't need a copy, I just need to append the whole array to the other one in place. How would you do this?

rubensworks: Oh ok, I misunderstood, nvm my comment then.

(some documentation would have helped here 😅)

jitsedesmet: You're right the function isn't that trivial when you didn't write it yourself. :)

lib/functions/Core.ts
jitsedesmet: I refactored this a bit. I hope it's better now.

lib/functions/OverloadNode.ts
```diff
@@ -0,0 +1,90 @@
...
+    }
+    this.subTrees[str]._addOverload(argumentTypes, func);
+  }
+
+  /**
+   * @param arg term to try and match to possible overloads of this node.
+   * @returns SearchStack a stack with top element the next node that should be asked for implementation or overload.
+   */
+  public getSubTreeWithArg(arg: E.TermExpression): SearchStack {
+    // The numbers here aren't magic.
```
rubensworks, line 46: This sentence can go :-)

lib/functions/OverloadNode.ts
```diff
@@ -0,0 +1,90 @@
...
+    // They just provide a priority, a way to tell what type is more concrete and should be looked at first.
```
rubensworks, line 47: Could you enumerate these types? Because it's not clear at the moment what these types are, and why you should look at some of them first.

rubensworks (approved):
Looks much clearer now :-)

Let's wait for @wschella's feedback, as this change is pretty invasive.

lib/functions/OverloadNode.ts
jitsedesmet: What do you mean?
Also I wouldn't give to much attention to these numbers as they should disappear the moment we implement the auto cast feature.

rubensworks: I'm referring to this:
> what type is more concrete

jitsedesmet: You mean just in the comment? 
Something like `The numbers a priority, a way to tell what type is more concrete and should be looked at first. Order of concreteness: 1. the type (eg: 'string'), 2. termType (eg: literal) and 3. a term.`

rubensworks: Yes, something like that indeed.
But perhaps you could include an `@see` reference to a place where the reader can find more information about what type, termType and term mean in the context of sparqlee (because I don't fully understand it atm).

lib/functions/OverloadNode.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * @param arg term to try and match to possible overloads of this node.
+   * @returns SearchStack a stack with top element the next node that should be asked for implementation or overload.
+   */
+  public getSubTreeWithArg(arg: E.TermExpression): SearchStack {
+    // The numbers a priority, a way to tell what type is more concrete and should be looked at first.
+    const matching: [number, OverloadNode][] = [];
+    for (const [ type, node ] of Object.entries(this.subTrees)) {
```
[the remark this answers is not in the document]
wschella: For later (with extensionMap and decent suptyping and shit): we want to prioritize runtime performance over 'compile time' performance (read initialization). We could keep a sorted list of valid keys (in parallel with the subtrees), and then we only need to loop over that in order. No more sorting at runtime needed.

This would also be okay to do already, but a maybe bit overkill if you can do it with 3 if statements (up to you).

wschella (changes requested):

lib/functions/Core.ts
```diff
@@ -1,32 +1,26 @@
-// eslint-disable-next-line no-redeclare
-import type { Map } from 'immutable';
-import { List } from 'immutable';
-
 import type * as E from '../expressions';
 import type * as C from '../util/Consts';
 import * as Err from '../util/Errors';
+import type { OverloadNode } from './OverloadNode';
```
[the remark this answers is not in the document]
jitsedesmet: I changed it, when we call it a tree I also think we should allow fewer functions so I made some functions private.

lib/functions/OverloadNode.ts
```diff
@@ -0,0 +1,89 @@
...
+    this._addOverload([ ...argumentTypes ], func);
+  }
+
+  private _addOverload(argumentTypes: ArgumentType[], func: E.SimpleApplication): void {
+    const argument = argumentTypes.shift();
+    if (!argument) {
+      this.implementation = func;
+      return;
+    }
+    const str = argument;
```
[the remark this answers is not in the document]
jitsedesmet: Yes, that's some history thing. This can be changed.

lib/functions/Helpers.ts
```diff
@@ -34,16 +49,19 @@ export class Builder {
...
-    const types = List(argTypes);
+    // TODO: Does this need to be a copy? - ask in PR.
+    const types = [ ...argTypes ];
     return this.add(new Impl({ types, func }));
   }
 
   public copy({ from, to }: { from: ArgumentType[]; to: ArgumentType[] }): Builder {
     const last = this.implementations.length - 1;
-    const _from = List(from);
+    // TODO: Does this need to be a copy? - ask in PR.
```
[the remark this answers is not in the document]
jitsedesmet: I'll leave rewriting this file for another time. I removed the TODO's :) Thank you!

wschella (approved):

lib/functions/OverloadNode.ts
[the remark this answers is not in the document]
jitsedesmet: I don't know whether we could work with an initialize phase. We'll have to see about that. For now I changed it to the if statements you asked for..

Issue #102, opened 2021-08-11
Auto casting arguments using `OverloadTree`


jitsedesmet:
With the newly added (#101 ) `OverloadTree` we can enable auto casting like talked about in #94. I create this new issue because #94  is getting a little big and would get much bigger. I'll link some of the, in my opinion, most relevant comments for this issue from #94  here.
* Auto casting idea: https://github.com/comunica/sparqlee/issues/94#issuecomment-895039787 and https://github.com/comunica/sparqlee/issues/94#issuecomment-895044029
* Needing depth first search: https://github.com/comunica/sparqlee/issues/94#issuecomment-895071748
* There's some talk about what types should be handled by auto casting. A clear scheme should be created for this. Once the scheme is ready I will link it in here.
* Rewriting builder: https://github.com/comunica/sparqlee/pull/101#discussion_r685864953

When using auto cast, it could happen that searching an overload operation is pretty expensive. The chosen overload for a given array of arguments will never change however. Each node could have some kind of cache to keep the already calculated `searchStack` in. This cache would make sparqlee have a global state. We want to take control over that state so when implementing the cache, we should provide functions to reset and maybe set the cache status. A package that could be used for this is: https://www.npmjs.com/package/lru-cache

If I forgot something, please add it in the comments. The issue became kinda big to keep track of :sweat_smile:

rubensworks:
👍 

> This cache would make sparqlee have a global state. We want to take control over that state so when implementing the cache, we should provide functions to reset and maybe set the cache status. A package that could be used for this is: https://www.npmjs.com/package/lru-cache

For this, I would recommend allowing this cache to be retrieved from the evaluator, so it can be reused in other evaluators by passing it into `*EvaluatorConfig`.

wschella:
If we can use a cache, why don't we just precompute? If we do the cache with `ArgumentType[]` as keys, we're back at the OverloadMap that was not a tree but a HashMap (and you need array equality again), with the difference being that the map now might incomplete (and thus smaller).

So the question/tradeoff for me is: you can generate this 'cache' in advance for all types, but it's a combinatorial explosion of entries in the map (i.e. use the tree only as an intermediate representation, and flatten it).  Might that be worth it (a bit more memory, but O(1) lookups)?  I don't really know. The cache might hit a good sweet spot.

[...]

Issue #87, opened 2021-07-29
Suport overriding functions


jitsedesmet:
Just like we recently added support for extension functions #16 we should also add support for overriding existing/ default named functions.

jitsedesmet:
I have an idea on how this needs to work:

To the config you provide a JS object with the mapping `named function -> overloadArguments -> function implementation`.
We could add this overload to the `OverloadTree` of that function. Having this behavior would make it extremely easy for users using there own types to add overloads for the named function which are mostly/ all casting functions. This is really good because it would provide an easy way to add casting to your own types. We know this in many OOP languages as well with the almost basic toString function.

The caviat in this implementation is that right now the `OverloadTrees` are global. This behavior requires them to be linked/ a field in an evaluator. In other words, each evaluator would need his own set of `OverloadTrees`. I don't think it would be to bad to have this and I'm pretty certain we also need this for #109 .

Implementing this should not drop performance. I think of implementing a hybrid solution. We keep the global `OverloadTrees` since we don't want to copy them for every evaluator when it's not needed. I would only make a (deep) copy of the trees when overriding functions are provided to the config of the evaluator.

jitsedesmet:
Another important note on this. This would mean all override functions are sync! At least as of right now.

rubensworks:
As this should be seen as a nice-to-have feature, and will probably be used only very rarely, I would not overcomplicate things.
I would suggest to just do the same as `extensionFunctionCreator`, and handle it similarly.

jitsedesmet:
I think there's a big difference to be made. The extension functions had no overloading. Named functions however, do have overloading. For this overloading we need the `OverloadTree`. 

If you're concerned about this issue getting really big or something, I don't think it will. I checked and I think the addition of these copies will be quite straight forward. Like I said, I think we need the copies for #109 . After we can make the copies, implementation of the `additional overrides for named operators` which is what this issue actually is, will be trivial.

jitsedesmet:
So if you say that you like the access point I suggest. (Adding the object `named function -> string[] -> function implementation`) I will get this done pretty soon.

rubensworks:
> The extension functions had no overloading.

But they _could_ have, right?

To avoid confusion, I would make sure extension functions and overrides are aligned.

If you think effort for all of this is low, go ahead :-)

jitsedesmet:
Well, pretty soon, I would first need to resolve #109 because this would require open world types within the `OverloadTree`.

> But they *could* have, right?

[...]

Issue #13, opened 2018-11-18
Implement Subtype Substitution 


wschella:
See [https://www.w3.org/TR/xpath-31/#dt-subtype-substitution](https://www.w3.org/TR/xpath-31/#dt-subtype-substitution), mostly relevant for the operator mapping.

wschella:
This is mostly implemented, although the docs should explain the workings some more.

Issue #28, opened 2019-05-16
Decimal formatting


wschella:
See https://github.com/w3c/rdf-tests/issues/58.

wschella:
The conclusion there seemed to be that we can output in whatever format we want, and the tests should allow both trailing and non-trailing forms.

We will choose non-trailing (as it is the default JS), and it makes more sense to me personally.

We will fix our local tests to allow this.

PR #103, opened 2021-08-11
Feature/auto casting function arguments

### LINKED ISSUE #12: Implement Type Promotion
See [https://www.w3.org/TR/xpath-31/#promotion](https://www.w3.org/TR/xpath-31/#promotion), mostly relevant for the arithmetic operators.

jitsedesmet:
This pull request aims to provide an implementation on auto casting functions as talked about in #102 . 
This pull request should *NOT* be merged before the scheme talked about in #102 is added.

There are a few questions to be answered before merging this PR too, those questions should be marked with a `TODO`.

Edit:
We can close a few issues with this PR: #13, #94 and #103.
When merging we open issues about: cast decimal so we can close #12 , questions surrounding `arithmeticWidening`, revision of https://github.com/comunica/sparqlee/pull/103#discussion_r687755303 (I think this issue is mainly for @wschella) and an issue surrounding caching and open world types.
The PR also makes implementation of #87 a lot easier.


lib/functions/OverloadTree.ts
```diff
@@ -1,18 +1,147 @@
 import type * as E from '../expressions';
+import type { LiteralTypes } from '../util/Consts';
+import { TypeAlias, TypeURL } from '../util/Consts';
 import type { ArgumentType } from './Core';
 
 export type SearchStack = OverloadTree[];
+export type OverRideType = LiteralTypes | 'term';
```
rubensworks, line 7: OverrideType

lib/functions/OverloadTree.ts
```diff
@@ -1,18 +1,147 @@
...
+
+  // Duration types
+  [TypeURL.XSD_DAYTIME_DURATION]: TypeURL.XSD_DURATION,
+  [TypeURL.XSD_YEAR_MONTH_DURATION]: TypeURL.XSD_DURATION,
+
+  // Stringly types
+  [TypeURL.RDF_LANG_STRING]: TypeAlias.SPARQL_STRINGLY,
+  [TypeURL.XSD_STRING]: TypeAlias.SPARQL_STRINGLY,
+
+  // XSD_ANY_URI is a weird one, it was handled equal to XSD_STRING before, we keep this? TODO - need to edit chart
```
rubensworks, line 36: I think that's correct indeed.

rubensworks (commented):
Given the size of this PR, I'm a bit scared of the lack of tests related to this.
We did indeed have tests already that will now cover part of this implementation, but looking at the coveralls output, it seems like coverage is not complete.

Therefore, I would recommend to add more detailed unit tests to the parts of the code you touched. Should be straightforward to do, now that this code is still fresh.

jitsedesmet:
> Therefore, I would recommend to add more detailed unit tests to the parts of the code you touched. Should be straightforward to do, now that this code is still fresh.

Do you think about changing the test directory structure at the same time? Like talked about in #34 ? So we can create a dir for unit tests as well?

lib/functions/OverloadTree.ts
jitsedesmet: Yes, after reading https://github.com/comunica/sparqlee/issues/102#issuecomment-896876623 and in particular https://www.w3.org/TR/xpath-31/#promotion it seems like that is the right behavior. I will add the reference to the code.

rubensworks:
> Do you think about changing the test directory structure at the same time? Like talked about in #34 ? So we can create a dir for unit tests as well?

Oh, I thought we did that one already 😅 

I think that's quite urgent, yes. But I would do it in a separate PR.

lib/functions/OverloadTree.ts
jitsedesmet: Taking this into account I think we should also allow the numeric types in this PR. https://github.com/comunica/sparqlee/issues/102#issuecomment-897103472

wschella:
I can review the PR this afternoon. I think changing the test folder structure can be done & and merged quickly. Then this PR can be rebased on it  (since it will be a significant one, and I do really propose fixing everything mentioned in #102 at once here, it's too closely related). We should indeed strive to not decrease test coverage here, but everything should be able to be tested in theory with some entries in a `TestTable`  somewhere.

lib/functions/OverloadTree.ts
jitsedesmet: Seems like it is not correct. Please take a look at https://github.com/comunica/sparqlee/issues/102#issuecomment-897427439 .

lib/functions/OverloadTree.ts
```diff
@@ -54,23 +77,49 @@ export class OverloadTree {
...
+    conversionFunction: (arg: E.TermExpression) => E.TermExpression, argumentTypes: ArgumentType[]): void {
+    if (!this.subTrees[typeToPromote]) {
+      this.subTrees[typeToPromote] = new OverloadTree(this.depth + 1);
+    }
+    this.subTrees[typeToPromote]._addOverload(argumentTypes, args => func([
+      ...args.slice(0, this.depth),
+      conversionFunction(args[this.depth]),
+      ...args.slice(this.depth + 1, args.length),
+    ]));
   }
```
jitsedesmet, line 116: This code should allow type promotion as talked about in https://www.w3.org/TR/xpath-31/#promotion . At least, how I understand it. If this is wrong please let me know.

rubensworks: At first glance it looks like it could work.
But we'll now for certain once we have tests covering these cases.

jitsedesmet: Yes, I will add the scheme first and then start writing the tests. With 'wrong' I meant the interpretation of 'promotion'. :)

README.md
```diff
@@ -325,17 +325,31 @@ When you create one, the SPARQL Algebra expression that is passed will be transf
...
-
-The type system is tailored for doing (supposedly) quick evaluation of overloaded functions. A function object consists of a map of argument types to a concrete function implementation for those argument types.
-
-When a function object is called with some functions, it looks up a concrete implementation. If we can not find one, we consider the argument of invalid types.
-
-Since many derived types exist, we also associate every literal with it's primitive datatype when constructing a literal. This handles **subtype substitution**, as we define allowed types in function of these primitives types. Note: the derived type is not maintained after an operation on the term, since I found no functions for which this was relevant.
-
-**Type promotion** is handled in a couple of ways. Firstly, a bit like C++ vtables, if we can not find a implementation for the concrete (primitive) types, we try to find an implementation for the term-types of all the arguments, if that fails, we look for an implementation on generic terms.
-
-We also handle type promotion by explicitly coding for it. This is done in the arithmetic functions `+, -, *, /`.
```
jitsedesmet, line 338: @wschella, the things that I removed here, what do they mean? 
Anyway, I should take a look at this!

lib/functions/OverloadTree.ts
```diff
@@ -8,11 +12,28 @@ export type SearchStack = OverloadTree[];
...
-  public constructor() {
+  public constructor(depth?: number) {
     this.implementation = undefined;
-    this.subTrees = {};
+    this.subTrees = Object.create(null);
+    this.depth = depth || 0;
+  }
+
+  /**
+   * Get the implementation that exactly matches @param args .
```
wschella, line 25:
```suggestion
   * Get the implementation for the types that exactly match @param args .
```

lib/functions/RegularFunctions.ts
```diff
@@ -341,17 +338,14 @@ const STRSTARTS = {
 
 const STRENDS = {
   arity: 2,
+  // TODO: What about (string, langstring)?
```
wschella, line 341: See above

lib/util/Consts.ts
```diff
@@ -16,149 +22,68 @@ export enum TypeURL {
...
-  double: TypeURL.XSD_DOUBLE,
-  decimal: TypeURL.XSD_DECIMAL,
-};
-
-export function decategorize(cat: PrimitiveNumericType): TypeURL {
-  return _decategorize[cat];
+  if ([ null, undefined, '' ].includes(typeString)) {
+    return TypeURL.XSD_STRING;
+  }
+  return TypeAlias.SPARQL_OTHER;
```
wschella, line 80: We should be able to get rid of this function in it's entirety (see discussion on type-attribute). Or am I missing something?

lib/util/TypeHandling.ts
```diff
@@ -0,0 +1,176 @@
+import type { LiteralTypes } from './Consts';
+import { TypeAlias, TypeURL } from './Consts';
+
+export type OverrideType = LiteralTypes | 'term';
+/**
```
wschella, line 5:
```suggestion
export type OverrideType = LiteralTypes | 'term';

/**
```

test/integration/functions/onStrings.test.ts
```diff
@@ -58,6 +60,7 @@ describe('string functions', () => {
       "👨‍👩‍👧‍👦"@en 2 = "‍👩‍👧‍👦"@en
       "👪"@en 1 1 = "👪"@en
       "👨‍👩‍👧‍👦"@en 1 1 = "👨"@en
+      "3"^^xsd:anyURI 1 1 = "3"
```
wschella, line 63:
```suggestion
      "type promotion"^^xsd:anyURI 2 4 = "ype"
```

type-scheme.svg
```diff
@@ -0,0 +1,931 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
wschella, line 1: We should move this to some `docs/` folder.

wschella (changes requested):
Very nice work, very important, and cleans things up nicely. 

I also think we we can make it a bit cleaner still: see comments and suggested changes.

lib/Transformation.ts
```diff
@@ -92,13 +97,23 @@ export function transformLiteral(lit: RDF.Literal): E.Literal<any> {
...
-      return new E.StringLiteral(lit.value);
+      return new E.StringLiteral(lit.value, lit.datatype);
+
     case DT.RDF_LANG_STRING:
       return new E.LangStringLiteral(lit.value, lit.language);
 
-    case DT.XSD_DATE_TIME:
-    case DT.XSD_DATE: {
+    case DT.XSD_DATE_TIME_STAMP:
+    case DT.XSD_DATE_TIME: {
```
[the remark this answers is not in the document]
jitsedesmet: You're right, keen eye!

lib/util/Consts.ts
```diff
@@ -3,11 +3,17 @@ import type * as RDF from 'rdf-js';
...
 
-export const TRUE_STR = '"true"^^xsd:boolean';
-export const FALSE_STR = '"false"^^xsd:boolean';
-export const EVB_ERR_STR = '"not an dateTime"^^xsd:dateTime';
+export type LiteralTypes = TypeAlias | TypeURL;
+
+export enum TypeAlias {
+  // Numeric is everything defined in https://www.w3.org/TR/sparql11-query/#operandDataTypes
+  SPARQL_NUMERIC = 'SPARQL_NUMERIC',
+  SPARQL_STRINGLY = 'SPARQL_STRINGLY',
```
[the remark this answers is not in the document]
jitsedesmet: Do we have one for this? I'd need to check it. I thought you suggested creating this type for some cleaner code?

lib/expressions/Term.ts
```diff
@@ -65,7 +68,9 @@ export class BlankNode extends Term {
 // Literals-- -----------------------------------------------------------------
 export class Literal<T> extends Term {
   public termType: 'literal' = 'literal';
-  public type: C.Type;
+  // TODO: Ask for PR: do we still need this type? I keep it just for NonLexicalLiteral right now.
+  //  Do we need the TypeAlias.SPARQL_NON_LEXICAL? We might be able to just use typeURL.value?
+  public type: LiteralTypes;
```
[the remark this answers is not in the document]
jitsedesmet: Awesome, removing that would provided some even cleaner code. Personally I would go for `dataType` because I think `type` feels kind of empty. (It doesn't mean much)

lib/expressions/Term.ts
```diff
@@ -91,38 +96,57 @@ export class Literal<T> extends Term {
...
+
+        return `${mantisse}E${exponent}`;
+      };
+      // // Be consistent with float
+      // decimal: (value) => {
+      //   const jsDecimal = value.toString();
+      //   return jsDecimal.match(/\./)
+      //     ? jsDecimal
+      //     : jsDecimal + '.0';
+      // },
```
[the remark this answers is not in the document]
jitsedesmet: Right, about this, what doe it mean? It's kind of vague for an outsider, could you elaborate on it? Then I will also have a clearer view on what exactly to do with it.

lib/expressions/Term.ts
```diff
@@ -91,38 +96,57 @@ export class Literal<T> extends Term {
...
+    }
+  }
+
+  // ExtensionTable[type][TypeAlias.SPARQL_NUMERIC] !== undefined
```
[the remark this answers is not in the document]
jitsedesmet: Yes, that's not clear at all. I wanted to have some way of telling this type was a `SPARQL_NUMERIC` but I ended up checking this in the constructor so it can go.

lib/functions/OverloadTree.ts
```diff
@@ -54,23 +77,49 @@ export class OverloadTree {
...
       this.implementation = func;
       return;
     }
     if (!this.subTrees[argumentType]) {
-      this.subTrees[argumentType] = new OverloadTree();
+      this.subTrees[argumentType] = new OverloadTree(this.depth + 1);
+    }
+    this.subTrees[argumentType]._addOverload(_argumentTypes, func);
+    // Defined by https://www.w3.org/TR/xpath-31/#promotion .
+    //  When a function takes a string, it can also accept a XSD_ANY_URI if it is cased first.
```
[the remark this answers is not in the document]
jitsedesmet: I would want to say 'to cast something in the past tense' but I have no idea if it is spelled correctly.

lib/util/Consts.ts
jitsedesmet: If by `this function` you mean `decatogorize`, it's gone?

test/integration/functions/onStrings.test.ts
```diff
@@ -20,6 +20,7 @@ describe('string functions', () => {
   });
 
   // TODO: Add errors for when non BCP47 strings are passed
+  //  this is equivalent to making argument of langMatches XSD_language?
```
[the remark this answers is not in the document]
jitsedesmet: What do you mean?

lib/util/Consts.ts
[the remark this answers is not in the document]
rubensworks: Yes, this definitely needs some kind of description, because this doesn't appear to be part of the SPARQL/RDF spec (?).

test/integration/functions/onStrings.test.ts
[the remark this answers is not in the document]
rubensworks: In RDF, strings with a language tag can not have a datatype, and vice-versa.

type-scheme.svg
rubensworks: Nice figure btw :-)

lib/util/Consts.ts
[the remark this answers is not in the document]
wschella: I did suggest it indeed, and it is present in the spec, but it is hidden a bit (can't put all the details up front ofc).

The first paragraphs of https://www.w3.org/TR/sparql11-query/#func-strings provide some more context. There is a separate section on argument compatibility that is relevant for some other issues as well. Let's summarize this together with the link and add it to the code.

lib/functions/OverloadTree.ts
wschella: A special thing to consider here is that subtype substitution and type promotion can work at the same time. There certainly should be tests for that, as I'm not sure the code handles this like it is (but it's also likely that I'm just missing something).

README.md
wschella: Resolved in suggestion.

lib/util/Consts.ts
```diff
@@ -3,11 +3,17 @@ import type * as RDF from 'rdf-js';
...
+  SPARQL_OTHER = 'SPARQL_OTHER',
+  SPARQL_NON_LEXICAL = 'SPARQL_NON_LEXICAL',
+}
 
-// TODO: Consider inlining all with 'const enum'
+// TODO: Consider inlining all with 'const enum' - do we need to handle this right now? Can this be removed?
```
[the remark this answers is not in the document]
wschella: `const enum` is a TS feature https://www.typescriptlang.org/docs/handbook/enums.html#const-enums that resolves some enum handling at compile time vs runtime. It might be worth to use these for speed, it might not be for other reasons, but we should at least write down why.

lib/expressions/Term.ts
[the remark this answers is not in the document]
wschella: Agreed.

lib/expressions/Term.ts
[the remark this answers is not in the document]
wschella: The SPARQL spec defines a way to format floats, the JS spec does the same. These aren't necessarily compatible. I can't recall why this is just commented out (it's just adds a trailing `.0` if the number is round), but it might be hint we don't have sufficient tests yet that the way we format things agrees with the SPARQL spec (because we use JS defaults).

lib/expressions/Term.ts
```diff
@@ -91,38 +96,57 @@ export class Literal<T> extends Term {
...
-  };
-
-  public type: C.PrimitiveNumericType;
+    if (typeCanBeProvidedTo(type, TypeURL.XSD_INTEGER)) {
+      this.specificFormatter = value => value.toFixed(0);
+    } else if (typeCanBeProvidedTo(type, TypeURL.XSD_DECIMAL)) {
+      this.specificFormatter = value => value.toString();
+    } else if (typeCanBeProvidedTo(type, TypeURL.XSD_FLOAT)) {
+      this.specificFormatter = value => value.toString();
+    } else if (typeCanBeProvidedTo(type, TypeURL.XSD_DOUBLE)) {
```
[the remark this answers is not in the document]
wschella: The problem here is mostly runtime overhead I think. These are a lot of checks we need to for each number, which is a very common datatype.
If we have a subtypesFor, we can create a static map matching each concrete subtype to a concrete formatter a bit faster. Not sure if worth, just a gut feeling.

lib/functions/OverloadTree.ts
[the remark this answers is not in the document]
wschella: it's just `if it's cast first` then

README.md
jitsedesmet: I unresolved this since It's still imported for me to check the functions and I'll use this as a reminder :)

lib/Transformation.ts
```diff
@@ -76,6 +76,11 @@ function transformWildcard(term: Alg.WildcardExpression): E.Expression {
...
 
 // TODO: Maybe do this with a map?
+/**
+ * Some of the types here have some super relation.
+ * This should match the relations provided in @see{extensionTableInput}.
+ * @param lit
+ */
 export function transformLiteral(lit: RDF.Literal): E.Literal<any> {
   if (!lit.datatype) {
     return lit.language ?
```
jitsedesmet, line 86: I was wondering if we could get some kind of reference on why we do this? It actually reminds me a bit of #14 ? Anyway, it's peculiar and I think we should write down what happens here.

jitsedesmet: Same thing for the cases of null, undefined and '' underneath this.

lib/expressions/Term.ts
```diff
@@ -91,38 +96,57 @@ export class Literal<T> extends Term {
 }
 
 export class NumericLiteral extends Literal<number> {
-  private static readonly specificFormatters: {[key in C.PrimitiveNumericType]: (val: number) => string } = {
+  private readonly specificFormatter: ((val: number) => string);
+  public constructor(
+    public typedValue: number,
+    public typeURL: RDF.NamedNode,
+    public strValue?: string,
+    public language?: string,
```
jitsedesmet, line 104: Am I wrong to think we want to make strValue mandatory? If you give a literal to sparqlee and ask it back immediately. Shouldn't we get the same thing? This is particularly interesting for all types date related. If you where to give a date without a time zone, you wouldn't want to get one back with a arbitrary timezone? Wouldn't it pay of to generalize this?

lib/util/Consts.ts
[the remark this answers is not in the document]
jitsedesmet: > take a string literal as an argument and accept a simple literal, a plain literal with language tag, or a literal with datatype xsd:string
This means a simple literal. Whatever that means, is also a sparql_stringly. We should make sure this is True as well.

lib/util/Consts.ts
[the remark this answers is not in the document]
jitsedesmet: I'm thinking we don't want this? We probably want to add an extra option to the evaluatorConfig called `additionalExtentionTable` or smth. If we were to add this we need the runtime values. 
It's possible I don't understand const enum  enough but I think it would remove the runtime values.

lib/expressions/Term.ts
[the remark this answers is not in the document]
jitsedesmet: I agree, that's also a reason I commented https://github.com/comunica/sparqlee/pull/103#discussion_r688370200 .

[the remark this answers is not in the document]
jitsedesmet: We might need to think about creating js Classses representing the various datatypes? I don't know if this would break things, if it would be a lot of work or if it's even possible. I do think it can be interesting for formatting purposes. In addition, our TypeUrl currently only extend 1 thing. That means we could map this perfectly to js class extension? Wouldn't that be great?

lib/Transformation.ts
wschella: You mean the fact that if there's no datatype this is considered a string?

jitsedesmet: Yes.
Also in https://www.w3.org/TR/sparql11-query/#func-lang I found
> Note that the RDF data model does not include literals with an empty language tag.

Does this have anything to do with this?

wschella: It's this one again: https://www.w3.org/TR/sparql11-query/#func-string.

wschella: Or at least largely. I don't think there are any functions that take an xsd:string but not a simple literal (no datatype).

wschella: I think there has been some discussion in IDLab regarding this as well, and possible spec incompatibilities. Maybe @rubensworks has some better references.

wschella: See also https://www.w3.org/TR/2004/REC-rdf-concepts-20040210/#dfn-plain-literal .

The distinction between simple literal and plain literal has never been clear to be, as plain literals are simple literals but with an (optional) language tag. So a plain one without a language tag is just a simple one?

The key part is simple, explicitly typed, and language fall under "string literal" (i.e. stringly in our case), and there are IIRC functions that take an explicitly typed xsd:string but not a simple (no datatype, no language) one.

There are functions that take xsd:string + simple, but don't take languageStrings.

rubensworks: This issue should contain all the answers :-) https://github.com/w3c/sparql-12/issues/112

lib/util/Consts.ts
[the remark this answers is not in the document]
wschella: This true because we consider simple literal (no datatype) to be xsd:string (see also https://github.com/comunica/sparqlee/pull/103#discussion_r688454246).

lib/util/Consts.ts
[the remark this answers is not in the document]
wschella: This seems likely.

lib/util/Consts.ts
wschella: Something weird happened, I meant the `type(typeString: string): LiteralTypes` function.

lib/Transformation.ts
wschella: Ah exactly. And for reference, we already do RDF 1.1 behavior. If there is no datatype, we make it xsd:string. (except when there is a language tag ofc)

lib/expressions/Term.ts
wschella: This is optional because we don't format prematurely when calculating, as the value could be changed many times, and we only care for formatting when exporting.

If you simply pass a literal from somewhere else and request that, strValue should be populated, and will be returned. See the `str` function. 

Anyway, this is only for performance, you need to have correct formatting in any case.

Datetime is an exception because there are no functions that manipulate the (typed) value https://www.w3.org/TR/sparql11-query/#func-date-time .

Might be worth adding this to some docs.

lib/expressions/Term.ts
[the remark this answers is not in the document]
jitsedesmet: Those are indeed a lot of checks for every number. would it help to make a function that would do these checks only when needed in `str()` ?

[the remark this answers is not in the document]
wschella: It's better. But I still think a static map is the ideal approach here. Your choice ofc. You can also split the numeric class into the primitives (float, decimal, double), but that won't solve much.

You need either (derived type) -> (primitive type) -> (formatter), or (derived type) -> (formatter). There's plenty of options for building one I think.

test/integration/functions/onStrings.test.ts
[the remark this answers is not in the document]
jitsedesmet: Oh I see, alright. Thanks

lib/util/Consts.ts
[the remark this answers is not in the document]
jitsedesmet: Since I stumbled upon this problem like 5 times now. Should we put it in the readme or would that be overkill?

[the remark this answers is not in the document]
jitsedesmet: I'll just write it very clearly in the places I had issues with it. If I put it in the readme the chance is low someone will read it.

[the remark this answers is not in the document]
wschella: I don't think it's necessary. Maybe mentioning simple literals don't exists because we make them xsd:string is worth it.

lib/util/Consts.ts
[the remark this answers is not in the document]
jitsedesmet: Okey, Should I keep the original `TODO` or can I remove it?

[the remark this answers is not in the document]
wschella: As you see fit.

lib/expressions/Term.ts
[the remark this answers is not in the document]
jitsedesmet: @rubensworks left me this: https://github.com/w3c/rdf-tests/issues/58  and that points to #28 . I think this refers to the code here? That would mean we can just remove the comment and there is no reason to handle it?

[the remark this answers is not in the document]
wschella: Makes sense, let's clean it up.

lib/expressions/Term.ts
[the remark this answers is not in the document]
jitsedesmet: I don't think we can do a static map. Especially if we move sparqlee to have an open world mentality. Having a list of all subtypes would not be possible I think. Or just not scalable.

And like you said, splitting this class will not solve much. In that case we will need to do the check at construction like it is now. I'm sorry, I'll need to disappoint you. I will create a `specificFormattrCreator`.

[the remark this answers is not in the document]
wschella: I don't think there is a scenario where the static maps don't scales but these checks do?

You are overestimating the likelihood of people providing _even more_ numeric subtypes, and sacrificing performance now for an unlikely scenario in a future world where sparqlee might be open world.

Even then, when it is open world, it would be preferable to have a static map, and a fallback mechanism for when the concrete type is not present.

lib/Transformation.ts
jitsedesmet: Alright, thank you for this reference. Now I will add this reference to the code. Should I add a counter as well? :p (jk)

lib/expressions/Term.ts
jitsedesmet: Alright, thanks for the clarification.

lib/util/Consts.ts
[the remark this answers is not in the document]
jitsedesmet: I think we can remove it.

lib/expressions/Term.ts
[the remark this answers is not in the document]
jitsedesmet: Maybe I just don't see it I'll make an issue about this and leave it with a creator?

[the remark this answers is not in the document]
wschella: That's okay :)

lib/functions/OverloadTree.ts
jitsedesmet: I pretty sure this wouldn't be a problem since promotion is always from a non super type. To another type. And of course like it is now, having a restriction of anyURI provided to a function expecting an xsd_string, the restriction will be substituted to anyUri and will then be promoted to xsd_string.

jitsedesmet:
When merging this PR we should also open 2 new issuees.
1. Is arithmetic widening the right choice? This PR creates a function `arithmetic widening` this function copy's  previous behavior. We should check if this behavior is spec compliant and add links to that part of the spec. Reason: it could be that short + short \ne short. Example: is 240 + 100 still a short? It can not be represented as a short. Should we check this? Or do we need to return a short with value 340 which is not a valid short value.
2. The requested change by @wschella right here https://github.com/comunica/sparqlee/pull/103#discussion_r687755303 . Should be checked closer. resolving that issue should also provide an explanation on any choices made there.

lib/util/Consts.ts
```diff
@@ -3,11 +3,17 @@ import type * as RDF from 'rdf-js';
...
-export const TRUE_STR = '"true"^^xsd:boolean';
-export const FALSE_STR = '"false"^^xsd:boolean';
-export const EVB_ERR_STR = '"not an dateTime"^^xsd:dateTime';
+export type LiteralTypes = TypeAlias | TypeURL;
+
+export enum TypeAlias {
+  // Numeric is everything defined in https://www.w3.org/TR/sparql11-query/#operandDataTypes
+  SPARQL_NUMERIC = 'SPARQL_NUMERIC',
+  SPARQL_STRINGLY = 'SPARQL_STRINGLY',
+  SPARQL_OTHER = 'SPARQL_OTHER',
```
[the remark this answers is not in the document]
jitsedesmet: I removed it.

lib/util/Consts.ts
```diff
@@ -3,11 +3,17 @@ import type * as RDF from 'rdf-js';
 
 const DF = new DataFactory();
 
-export const TRUE_STR = '"true"^^xsd:boolean';
-export const FALSE_STR = '"false"^^xsd:boolean';
-export const EVB_ERR_STR = '"not an dateTime"^^xsd:dateTime';
+export type LiteralTypes = TypeAlias | TypeURL;
```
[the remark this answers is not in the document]
jitsedesmet: What name do you suggest?

lib/expressions/Term.ts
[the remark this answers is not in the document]
jitsedesmet: I removed both `type` and `typeUrl` and replaced them by `dataType: string` and in certain Literals this becomes `dataType: LiteralTypes`. Since `LiteralTypes` is a subset of string I think this is a pretty solution.

lib/util/Consts.ts
jitsedesmet: Yes, I was able to remove this function.

README.md
jitsedesmet: This is done now. I checked all functions.

jitsedesmet:
If we wish to implement https://www.w3.org/TR/xpath-31/#promotion completely, we should make sure we cast the `decimal` to `double` or `float` when promoting `decimal`. I will leave this for now. I think we should open a new issue adding this (I think small) feature. I feel like we better do this in a new issue because we might need a package for this? I'm not sure though

lib/functions/OverloadTree.ts
jitsedesmet: This conversation has been resolved by type promotion.

lib/functions/OverloadTree.ts
jitsedesmet: I'm still convinced we can't really test this. So unless someone can give an example on when this could create a problem I consider this to be done.

jitsedesmet: I added some documentation so people don't break the system and also a test that first performs substitution and that promotion.

rubensworks (approved):
This stuff wasn't trivial, so really good work!

Before we merge, can you again double-check if all tests (unit, integration, spec, ...) still pass in Comunica?

Could you also list in the main description of this PR, what issues can be closed?

I'll wait a bit for merging until @wschella gives his 👍

jitsedesmet:
Thank you!

> Before we merge, can you again double-check if all tests (unit, integration, spec, ...) still pass in Comunica?

I checked the overal tests and ran test, integration and spec in the actor-init-sarql package. All tests succeeded.

I will add the closing issues.

rubensworks:
@wschella Merged this one already. Feel free to open issues in case you still see issues with this.

Issue #33, opened 2019-05-20
Benchmark performance


wschella:
We should benchmark the performance sometimes.

PR #106, opened 2021-08-17
Adding simple addition benchmark

jitsedesmet:
This pull request adds a simple benchmark that tests a simple addition of integers. This feature was requested  in #33 . Before the refactoring of #103 this test's output was: `bench addition x 1,595 ops/sec ±2.99% (83 runs sampled)`.
After refactoring we have `bench addition x 1,582 ops/sec ±2.04% (82 runs sampled)`. This means there isn't really a big performance loss.
A simple integer addition is probably one of the most expensive operations in the framework created in #102 so that's pretty good.

I use the benchmark lib that was already present and took most/ all of my inspiration from https://github.com/comunica/sparqlee/commit/760603210632a9c7ca2e1453611253f8ab1a0efa , provided to me by @wschella . For which I thank him.


rubensworks (approved):

wschella (changes requested):
~~benchemark.ts -> benchmark.ts~~ benchemark.ts -> integer-addition.ts
benchmark/ -> benchmarks/ (to more easily add extra benchmarks)

jitsedesmet:
I think we should be able to merge this now.

PR #110, opened 2021-08-18
Document sparql spec of arithmetic types

### LINKED ISSUE #107: Document spec link explaining `arithmeticWidening`
We should make sure we understand why we use `arithmeticWidening` and where this behaviour is mentioned in the specs. 
This issue has 2 tasks.
1. Find the specs surounding `arithmeticWidening`. 
2. Implement `arithmeticWidening` like the specs sugest.

`arithmeticWidening` answers a simple question: what type is the return type of an arithmetic operation. 
things to consider: it could be that short + short \ne short. Example: is 240 + 100 still a short? It can not be represented as a short. Should we check this? Or do we need to return a short with value 340 which is not a valid short value.

jitsedesmet:
This issue aim to close #107 . I documented some specs made changes where possible. I also removed the type widening functions since they were of no use anymore. And I don't think having them is all that beneficial.

I have already run the Comunica test suite over it as well, all looks well.


lib/functions/Helpers.ts
```diff
@@ -142,20 +141,27 @@ export class Builder {
   }
 
   /**
+   * !!! Be aware when using this function, it will create different overloads with diffrent return types !!!
```
rubensworks, line 144:
```suggestion
   * !!! Be aware when using this function, it will create different overloads with different return types !!!
```

rubensworks (approved):
LGTM!

PR #111, opened 2021-08-18
put ts in strict mode and fix

jitsedesmet:
This PR doesn't really fix any issues but it prevents them from occurring.
I enabled 
`
"strict": true,
"strictFunctionTypes": false,
"strictPropertyInitialization": false
`
in the typescript config and fixed the problems this caused.


lib/expressions/Term.ts
```diff
@@ -98,7 +98,7 @@ export class Literal<T> extends Term {
   }
 
   public str(): string {
-    // eslint-disable-next-line @typescript-eslint/no-base-to-string
+    // @ts-expect-error TS2339
```
rubensworks, line 101: Can we resolve this instead of ignoring it?

lib/functions/SpecialFunctions.ts
```diff
@@ -251,7 +251,11 @@ async function inRecursiveAsync(
   }
 
   try {
-    const next = await evaluate(args.shift(), mapping);
+    const nextExpression = args.shift();
+    if (!nextExpression) {
+      throw new Error('Create to catch');
```
rubensworks, line 256: This error message is a bit unclear (also the next one)

rubensworks (changes requested):
Good work! Just some minor comments.

lib/functions/SpecialFunctions.ts
jitsedesmet: There's a try catch. I'm fairly certain that when there's no nextExpression we should jump to the catch. This error would never see the surface. Maybe I should write that in the message?

lib/expressions/Term.ts
jitsedesmet: I don't think so. There are I think 3 choices here
1. it just returns the basic string (as it is now)
2. We create our own basic string and check if the type has .toString or not
3. We just throw some error (I don't see any reason to pick this one)

:)

lib/functions/SpecialFunctions.ts
rubensworks: I would advise against that, as it makes the logic flow harder to follow. Perhaps you can just `return` there instead?

lib/expressions/Term.ts
rubensworks: Looks like an issue with the generic type `T` then. It should probably enforce the type to have a `toString()` method.

lib/functions/SpecialFunctions.ts
jitsedesmet: That should be possible yes.

lib/expressions/Term.ts
jitsedesmet: Sure, I could solve that but that would make `NonLexicalLiteral` a little ugly.

jitsedesmet: I'll change it in the next push and then you can see whether you like it.

lib/functions/SpecialFunctions.ts
jitsedesmet: Wow, just realize we don't need any of that. We check the length so we could just use a cast.

jitsedesmet:
Btw, i would like to ask you not to merge this yet. Even when you think it's ready :)

jitsedesmet:
Alright, we should be able to merge this now.

lib/expressions/Term.ts
rubensworks: No changes?

PR #112, opened 2021-08-20
Feature/buffer & open world typing

### LINKED ISSUE #109: OverloadTree caching and open world types
These are 2 problems in 1 issue because they need to be implemented with much consideration with eachother.
There has already been a lot of talking surrounding the cache part of this issue in #102 . To make sparqlee truly open world I think we still need to add the functionality of a user providing his own datatypes and making them have the known datatypes as a super type. This way those types can still be provided to functions expecting xsd datatypes due to subtype substitution.

jitsedesmet:
This pull request aims to close #109 .

What this PR *does*:
 - Split `lib/transformation.ts` in 2 classes that are easier to read and maintain.
 - Split `lib/Aggragators` in multiple files and create a base class for the evaluators to use instead of calling global functions.
 - Create some more super classes for the Evaluator files that makes us handle them more like the same. (Reduces the need to copy everything from sync to async)
 - Remove `TypeCheckedLiteral`  
Reason: A literal almost always has a known dataType. Only when transforming an RDF.Term to it's internal representation does it have an other type. Every operation on a custom function return a known type. For example providing '-5 ' with dataType myNegatives extends Integer provided to UnaryMinus should return the known *known type* integer and not something with dataType myNegatives. Typechecking every internal literal would thus be an unnecessary overhead.
- Update `tests/util/generalEvaluation` I added a `generalErrorEvaluation` function. Using generalEvaluate in combination with an ErrorTable made SyncAvaluator not fully tested. When the `AsyncEvaluator` throws an error in an ErrorTable. We still want to test the `SyncEvaluator`. This function makes sure this happens by catching all thrown Errors.
- Provides access to a shared (between sync and async) functionConfig from within functions that use the `OverLoadTree`. This allows functions to access the `IOpenWorldEnabler`. It also allows us to make IRI and NOW regular functions.  
I couldn't move `BNODE` because the sync and async implementation is different. It might be interesting to move this in the future since we probably want the `AsyncEvaluator` to be able to get an async `SuperTypeDiscoverCallback`. This should be possible but I would leave it for another PR. This one is big enough as is.
- This PR will also close #109. It does this by requesting 3 extra arguments from the user: 
```ts
overloadCache?: LRUCache<string, ImplementationFunction | undefined>;
typeCache?: LRUCache<string, GeneralSubExtensionTable>;
superTypeDiscoverCallback?: (unknownType: string) => string;
```
It should be noted that the values of the caches are not to be used by the user.
Not providing a `typeCache` results in us making one our self. Not providing an `overloadCache` will **not** create one.


lib/functions/RegularFunctions.ts
```diff
@@ -315,23 +329,23 @@ const STRLEN = {
...
     .onTernaryTyped([ TypeURL.XSD_STRING, TypeURL.XSD_INTEGER, TypeURL.XSD_INTEGER ],
-      (source: string, startingLoc: number, length: number) =>
+      () => (source: string, startingLoc: number, length: number) =>
         string([ ...source ].slice(startingLoc - 1, length).join('')))
     .onTernary([ TypeURL.RDF_LANG_STRING, TypeURL.XSD_INTEGER, TypeURL.XSD_INTEGER ],
-      (source: E.LangStringLiteral, startingLoc: E.NumericLiteral, length: E.NumericLiteral) => {
+      () => (source: E.LangStringLiteral, startingLoc: E.NumericLiteral, length: E.NumericLiteral) => {
         const sub = [ ...source.typedValue ].slice(startingLoc.typedValue - 1, length.typedValue).join('');
         return langString(sub, source.language);
       })
```
jitsedesmet, line 351: Question here before I make an issue about it. I think this implementation is wrong. The third argument of slice is the endIndex. However the third argument of https://www.w3.org/TR/sparql11-query/#func-substr is the length of the substring.
This function is also based on https://www.w3.org/TR/xpath-functions/#func-substring . It talks about arguments being double and sparql talks about integers. Maybe we should provide both implementation? Doesn't hurt anyone?

rubensworks: >  I think this implementation is wrong.

That may be wrong indeed. Aren't there tests for this yet?

> It talks about arguments being double and sparql talks about integers. Maybe we should provide both implementation? Doesn't hurt anyone?

I think there is no need to do that. If I understand the XPath spec correctly, doubles are only allowed there for backwards-compatibility, which is a problem we don't really have here.

rubensworks (changes requested):
Looks good in general!

Just wondering I we shouldn't use the general config/context in places where the openworld config thing is passed.

lib/functions/Core.ts
```diff
@@ -1,9 +1,31 @@
...
+import type { Bindings } from '../Types';
 import type * as C from '../util/Consts';
 import * as Err from '../util/Errors';
-import type { OverloadTree } from './OverloadTree';
+import type { IOpenWorldEnabler } from '../util/TypeHandling';
+import type { ImplementationFunction, OverLoadCache, OverloadTree } from './OverloadTree';
 
-type Term = E.TermExpression;
+export interface IFunctionContext {
+  openWorldEnabler: IOpenWorldEnabler;
```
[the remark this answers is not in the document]
jitsedesmet: Sure but I think we'll remove this considering you also asked to just pass the whole context to functions.

lib/transformers/AlgebraTransformer.ts
```diff
@@ -0,0 +1,170 @@
...
+  private readonly overloadCache: OverLoadCache;
+  private readonly applyContext: IApplyFunctionContext;
+  public constructor(algebraConfig: AlgebraTransformConfig) {
+    super({ discoverer: algebraConfig.superTypeDiscoverCallback, cache: algebraConfig.typeCache });
+    this.funcContext = {
+      openWorldEnabler: this.openWorldType,
+      baseIRI: algebraConfig.baseIRI,
+      now: algebraConfig.now,
+    };
+    // @ts-expect-error TS2322
```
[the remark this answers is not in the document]
jitsedesmet: Replaced it by a cast :)

lib/util/TypeHandling.ts
```diff
@@ -120,31 +148,55 @@ export function isTypeAlias(type: string): TypeAlias | undefined {
...
-  if (isLiteralType(type) || type === 'term') {
+  if (isKnownLiteralType(type) || type === 'term') {
     return <OverrideType> type;
   }
   return undefined;
 }
 
+export type TypeCache = LRUCache<string, GeneralSubExtensionTable>;
+export type SuperTypeDiscoverCallback = (unknownType: string) => string;
+export interface IOpenWorldEnabler {
```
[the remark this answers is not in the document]
jitsedesmet: Sounds good to me!

lib/functions/RegularFunctions.ts
jitsedesmet: There are tests but they test with arguments someStr 1 1 which is gives the same result.
Alright. Should I edit this in this PR or create an issue handling this small change?

jitsedesmet:
> Just wondering I we shouldn't use the general config/context in places where the openworld config thing is passed.

We could do this for sure. We can then also make `SuperTypeCallback` async. To handle this we would then also need to provide Static create method so the creation can be async. Some problems will arise handling normal construction but we'll see about that.

lib/functions/RegularFunctions.ts
rubensworks: I would do it in a separate PR right after this one. (or at least create an issue now)

README.md
```diff
@@ -142,6 +154,25 @@ config.extensionFunctionCreator = (functionName: RDF.NamedNode) => {
 }
 ```
 
+### Overload function caching
+
+An overloadcache allows Sparqlee to cache the implementation of a function provided the argument types. 
+This cache is only used when provided to the context.
```
rubensworks, line 160: Also say that this is safe to reuse across different evaluators, and that manual modification is not recommended.

rubensworks: Also, do the benchmarks show that this actually helps performance?

README.md
```diff
@@ -142,6 +154,25 @@ config.extensionFunctionCreator = (functionName: RDF.NamedNode) => {
...
+
+### Super type discovery
+
+The `superTypeDiscoverCallback` allow a user to use custom types and define there super relationship to other types.
```
rubensworks, line 164:
```suggestion
The `superTypeDiscoverCallback` allow a user to use custom types and define their super relationship to other types.
```

rubensworks: Can you also explain why this is needed/useful?

README.md
rubensworks: If so, make sure to explain that this improves performance.

index.ts
```diff
@@ -1,6 +1,6 @@
 // TODO: this form is deprecated, we should not rename these types. Should change in next mayor update.
-export { AsyncEvaluator, IAsyncEvaluatorConfig as AsyncEvaluatorConfig } from './lib/evaluators/AsyncEvaluator';
-export { SyncEvaluator, ISyncEvaluatorConfig as SyncEvaluatorConfig } from './lib/evaluators/SyncEvaluator';
+export { AsyncEvaluator, IAsyncEvaluatorContext as AsyncEvaluatorConfig } from './lib/evaluators/AsyncEvaluator';
+export { SyncEvaluator, ISyncEvaluatorContext as SyncEvaluatorConfig } from './lib/evaluators/SyncEvaluator';
```
rubensworks, line 3: We need to remain backwards-compatible, so make sure to re-export them as `IAsyncEvaluatorConfig` and `ISyncEvaluatorConfig` as well, but add a deprecation notice (with TODO to remove in next major version).

test/integration/functions/op.addition.test.ts
```diff
@@ -28,6 +35,8 @@ describe('evaluation of \'+\' like', () => {
       NaN    NaN    = NaN
       NaN    anyNum = NaN
       anyNum NaN    = NaN
+      
+      "2"^^example:int "3"^^example:int = ${int('5')}
```
rubensworks, line 39: Even though we only do sync for now, this is actually _really_ powerful in itself, by just writing a tiny config extension.

test/integration/functions/op.addition.test.ts
```diff
@@ -8,6 +9,12 @@ describe('evaluation of \'+\' like', () => {
     operation: '+',
     aliases: numeric,
     notation: Notation.Infix,
+    config: {
+      type: 'sync',
+      config: {
+        superTypeDiscoverCallback: unknownType => TypeURL.XSD_INTEGER,
```
rubensworks, line 15: I know I already suggested another name before, but what do you think about this one?
```suggestion
        getSuperType: unknownType => TypeURL.XSD_INTEGER,
```

rubensworks (changes requested):
Good stuff, just some minor comments.

README.md
jitsedesmet: > Also, do the benchmarks show that this actually helps performance?

Don't know yet because I don't have the benchmark in this PR.
Maybe we should merge the benchmark PR first?

rubensworks: Sure, once you give the go-ahead, we can do that.

README.md
jitsedesmet: Sure! Because it's cool :)

index.ts
jitsedesmet: I will, Thank you for noticing.

jitsedesmet: I will, thank you!

test/integration/functions/op.addition.test.ts
jitsedesmet: Is this a question? Or are you just saying it looks good?

test/integration/functions/op.addition.test.ts
jitsedesmet: Yes, my bad. Seems like I didn't change this all the way.

jitsedesmet: Do you mean to just edit the variable name or also the type name?

test/integration/functions/op.addition.test.ts
rubensworks: No question, just a statement :-)

test/integration/functions/op.addition.test.ts
rubensworks: Just the key name is fine.

rubensworks: Actually, the type name can be simplified to `ISuperTypeCallback`

jitsedesmet: It isn't an interface name though. So I would keep `SuperTypeCallback`. I did change the keyname to `getSuperType`.

test/integration/functions/op.addition.test.ts
jitsedesmet: Okay, I'm glad you like it too.

README.md
jitsedesmet: I added a benchmark test and cleaned the benchmarks a bid. It seems like it's 3x as fast. (in this benchmark)

rubensworks: > It seems like it's 2x as fast. (in this benchmark)

Oh wow, that's quite significant! I didn't expect it to be _that_ much faster. But that's a good thing :-)

Make sure to mention this in the docs!

We'll probably want to exploit this as much as possible from within Comunica then, so that we reuse this cache whenever possible. Perhaps you could look into this next?

jitsedesmet: Alright. I did some digging and it turns out my initial benchmark introduced in https://github.com/comunica/sparqlee/pull/106 was quite wrong. I updated the benchmark to return a mean execution time and this gave some more insights. I also added some benchmarks on past commits to find out where the big difference in execution time came from. It seems like the new type system is to blame.

I will provide a list of branches with added benchmarks and their reports.

- [benchmark on commit before using the overload tree](https://github.com/jitsedesmet/sparqlee/tree/execution-time/benchmark-before-overload-tree) -> `bench addition no overloadCache x 18.58 ops/sec ±1.96% (49 runs sampled)
Mean execution time without cache 0.05381279502040815`
- [benchmark on commit after using overload tree but with old type system](https://github.com/jitsedesmet/sparqlee/tree/execution-time/benchmark-after-using-overload-tree) -> `bench addition no overloadCache x 34.60 ops/sec ±1.78% (60 runs sampled)
Mean execution time without cache 0.028901073775000004`
- [benchmark on commit using overload tree and new type system (with type substitution and type promotion)](https://github.com/jitsedesmet/sparqlee/tree/execution-time/benchmark-after-type-promotion-and-substitution) -> `bench addition no overloadCache x 7.45 ops/sec ±1.15% (23 runs sampled)
Mean execution time without cache 0.134191621826087`
- Benchmark on current PR -> `bench addition no overloadCache x 8.82 ops/sec ±1.46% (26 runs sampled)
bench addition with overloadCache x 22.57 ops/sec ±0.76% (41 runs sampled)
Mean execution time without cache 0.11341942315384619
Mean execution time with cache 0.044309969926829264
Fastest is bench addition with overloadCache`

From this we can conclude that the new type system is a lot slower then the old one. I already said that the integer addition is one of the hardest operations to perform on this system. Overall we can conclude that using a cache the new system is still faster than what we had before. I am personally really happy about this considering we have a *much* stronger type system that is fully spec complaint (ignoring the type promotion detail here).

A small thing to notice is how the overload tree is quite a bit faster than the Immutable list solution we had first.

Issue #119, opened 2021-08-27
Make new type system default


jitsedesmet:
The new type system has been put behind a flag in #117. We should try to make this new type system the default.
This means that we should optimize the new system some more. Even seemingly small optimization mean a lot here. The functions that require special attention are:
* `experimentalTransformLiteral` in lib/transformers/TermTransformer
* `isSubTypeOf` in lib/util/TypeHandling
* `getSubTreeWithArg` in lib/functions/OverloadTree

Of course we could also look at the few functions that are called within these functions.

I don't have any suggestions on how to start optimizing these functions. If there  are any more questions regarding this issue you should be able to contact me on them.

PR #120, opened 2021-08-27
Speed up new type system when using no cache

jitsedesmet:
This PR just removes the use of Object.entries in the `getSubTreeWithArg` function. This almost halves the execution time of the new type system when not using a cache or when the cache misses.

It can be seen as a part of #119 and might provide an example of the kind of optimizations we need.


rubensworks (commented):
So this improves performance at the cost of memory usage.

Makes me wonder if `this.subTrees` is still needed here (in its current form)?

PR #121, opened 2021-08-28
Execution time/increase transform literal execution time

jitsedesmet:
This PR builds on top of #120 . It increases the execution time to a point where we almost want to keep the new type system I think.

Execution times:
```
bench addition no overloadCache x 17.26 ops/sec ±2.23% (47 runs sampled)
bench addition with overloadCache x 26.19 ops/sec ±1.31% (46 runs sampled)
bench addition non experimental x 26.96 ops/sec ±1.34% (48 runs sampled)
Mean execution time without cache 0.05794374734042553
Mean execution time with cache 0.03818295654347826
Mean execution time with nonExperimental 0.037093097125
Fastest is bench addition non experimental
```

This means there is only a +-5% decrease when using the new type system.


lib/util/TypeHandling.ts
```diff
@@ -269,26 +269,29 @@ export function isInternalSubType(baseType: OverrideType, argumentType: KnownLit
...
+    return superTypeDictTable[concreteType];
+  }
+  // Datatype is a custom datatype
+  return getSuperTypes(baseType, superTypeProvider);
+}
+
 /**
  * This function needs do be O(1)! The execution time of this function is vital!
  * We define typeA isSubtypeOf typeA as true.
+ * When you find yourself calling this function a lot (e.g. in a case statement),
```
rubensworks, line 285: Perhaps it makes sense to remove this function even?

rubensworks (commented):
These changes give another reason why Sparqlee would benefit from being migrated to a proper object-oriented architecture.

wschella:
Should we rebase the relevant commits onto the master for this PR and the other similar ones?

jitsedesmet:
Yes, we should :) I plan on doing this before Monday but a lot is happening right now so it might get delayed. This PR and #120 should just get rebased and then they should be ready to merge. Like I said, we should first finish #120 and Ruben asked a question there that needs to be looked into.

rubensworks:
Great work!

PR #139, opened 2022-05-04
Returns a consistent order for uncomparable literals

Tpt:
Before, if the two compared literals had `typedValue`s for which comparison was not supported by JavaScript, the first argument was always considered as lower than the second.

In this pull request, we compare first the kind of literals using buckets ("number", "string"...). If the two literals are in the same bucket, we proceed to the actual comparison.
If not, we compare the literals using the bucket identifier.
We do that instead of comparing directly the datatypes to avoid problems like `"xsd:decimal"` < `"xsd:entity"` < `"xsd:integer"` even if some decimals should be lower than integers.

This PR also simplifies the ordering tests to make sure all combinations of argument order and ascending/descending expectation are tested.

This PR removes the public Ordering.order that was completely unused.

The diffs are more readable if read commit by commit.

This is a first step to solve comunica/comunica#892

### DIFF
--- a/lib/util/Ordering.ts
+++ b/lib/util/Ordering.ts
@@ -1,39 +1,36 @@
 import type * as RDF from '@rdfjs/types';
 import * as LRUCache from 'lru-cache';
-import type * as T from '../expressions/Term';
 import { TermTransformer } from '../transformers/TermTransformer';
 import type { ISuperTypeProvider, SuperTypeCallback, TypeCache } from './TypeHandling';
 
 // Determine the relative numerical order of the two given terms.
 export function orderTypes(litA: RDF.Term | undefined, litB: RDF.Term | undefined, isAscending: boolean,
   typeDiscoveryCallback?: SuperTypeCallback, typeCache?: TypeCache, enableExtendedXSDTypes?: boolean): -1 | 0 | 1 {
+  if (litA === undefined || litB === undefined || litA.equals(litB)) {
+    return 0;
+  }
+  return isLowerThan(litA, litB, typeDiscoveryCallback, typeCache, enableExtendedXSDTypes) === isAscending ? -1 : 1;
+}
+
+function isLowerThan(litA: RDF.Term, litB: RDF.Term,
+  typeDiscoveryCallback?: SuperTypeCallback, typeCache?: TypeCache, enableExtendedXSDTypes?: boolean): boolean {
   const openWorldType: ISuperTypeProvider = {
     discoverer: typeDiscoveryCallback || (() => 'term'),
     cache: typeCache || new LRUCache(),
   };
   const termTransformer = new TermTransformer(openWorldType, enableExtendedXSDTypes || false);
-  if (litA && litA.termType === 'Literal' && litB && litB.termType === 'Literal') {
+  if (litA.termType === 'Literal' && litB.termType === 'Literal') {
     const myLitA = termTransformer.transformLiteral(litA);
     const myLitB = termTransformer.transformLiteral(litB);
-    return order(myLitA, myLitB, isAscending);
+    const aType = getValueClassName(myLitA.typedValue);
+    const bType = getValueClassName(myLitB.typedValue);
+    if (aType !== bType) {
+      return aType < bType;
+    }
+    return myLitA.typedValue < myLitB.typedValue;
   }
-  return 0;
-}
-
-// Effective ordering
-export function order(orderA: T.Literal<any>, orderB: T.Literal<any>, isAscending: boolean): -1 | 0 | 1 {
-  // If the two values have not the same typedValue type and, hence, are not comparable,
-  // we compare them using the typedValue type name.
-  const aType = getValueClassName(orderA.typedValue);
-  const bType = getValueClassName(orderB.typedValue);
-  if (aType !== bType) {
-    return aType > bType === isAscending ? 1 : -1;
-  }
-
-  if (orderA.typedValue === orderB.typedValue) {
-    return 0;
-  }
-  return orderA.typedValue > orderB.typedValue === isAscending ? 1 : -1;
+  // Order is not defined yet for other terms.
+  return true;
 }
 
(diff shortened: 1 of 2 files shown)


lib/util/Ordering.ts
```diff
@@ -22,8 +22,21 @@ export function orderTypes(litA: RDF.Term | undefined, litB: RDF.Term | undefine
 
 // Effective ordering
 export function order(orderA: T.Literal<any>, orderB: T.Literal<any>, isAscending: boolean): -1 | 0 | 1 {
+  // If the two values have not the same typedValue type and, hence, are not comparable,
+  // we compare them using the typedValue type name.
+  const aType = getValueClassName(orderA.typedValue);
```
rubensworks, line 27: Not sure if we'll catch all of the type distinctions by looking at the class name here.
For example, langstrings and plain strings will both produce the `string` type here.

It might be safer to instead check the `dataType` field of the literal.

rubensworks (changes requested):
That was quick! 🙂 

The tests look perfect to me, just not sure about the way types are compared (see comment below).

lib/util/Ordering.ts
Tpt: Thank you for your review!

Yes, indeed, the current code leads to orders like "a"@en < "b" < "c"@en.

To implement `dataType` lookup, I would need to change the `order` function signature to get the `typeDiscoveryCallback` value in order to make sure to properly compare comparable data types like `xsd:integer` and `xsd:decimal`. This method does not seem to be used anywhere, is it fine if I make it private to be able to tweak it easily?

rubensworks: Indeed, that function is not used anywhere (AFAICS), so it's fine to make it private.

I'm not sure if invoking `typeDiscoveryCallback` will be needed though, because some preprocessing (using `typeDiscoveryCallback`) on the terms is being done here: https://github.com/comunica/sparqlee/blob/a556549f7d366ae67b1e7f2a21e8bc7799e3d9b3/lib/util/Ordering.ts#L16-L17
But I don't know the details of this logic, so don't take my word for it 🙂

Tpt: Thank you! Indeed, there might be something to be done by using the `T.Literal` subclasses ids. The `dataType` field in the `T.Literal` is the original datatype and not the resolved one. I'm going to try something and update the PR for review.

Tpt: I have updated the PR to sort first using explicit "buckets" based on the typed term class names.

lib/util/Ordering.ts
```diff
@@ -1,29 +1,57 @@
...
 
-// Effective ordering
-export function order(orderA: T.Literal<any>, orderB: T.Literal<any>, isAscending: boolean): -1 | 0 | 1 {
-  if (orderA.typedValue === orderB.typedValue) {
-    return 0;
-  }
-  return orderA.typedValue > orderB.typedValue === isAscending ? 1 : -1;
-}
+const _TYPE_BUCKETS = new Map<any, string>();
+_TYPE_BUCKETS.set(T.BooleanLiteral, 'boolean');
```
rubensworks, line 50: I would recommend indexing these buckets by `dataType` (there should be constants defined for these somewhere), to reducing coupling with our literal class implementation.

rubensworks: Also, instead of defining the type buckets manually here, I'm wondering if we can't reuse some definitions from elsewhere for that.
For example: https://github.com/comunica/sparqlee/blob/a4eaeacb42f384457f9cee64a5b258efc2ff8747/lib/util/TypeHandling.ts#L19

Perhaps @jitsedesmet or @wschella have some thoughts on this.

Tpt: Thank you! Indeed, `mainSparqlType` seems the perfect function for this use case.

Tpt: I am fraid the coupling wih literal class is slightly unavoidable because the ordering implement for each kind of literals depends on how the literals are parsed. A cleaner option might be to rely on the `Literal.mainSparqlType` attribute instead of the class names.

`mainSparqlType` does not easily solve the problem because it would require to run the `typeDiscoveryCallback` before to make sure we use the normalized datatype and not the base one.

rubensworks: > A cleaner option might be to rely on the Literal.mainSparqlType attribute instead of the class names.

Sounds good to me!

> mainSparqlType does not easily solve the problem because it would require to run the typeDiscoveryCallback before to make sure we use the normalized datatype and not the base one.

Not sure I fully understand. AFAICS, `typeDiscoveryCallback` has already been called. And if it isn't, can't we invoke it directly here?

jitsedesmet: Oh, the scheme for this was /very/ easy to find since it is the only document in the `docs` folder.  https://github.com/comunica/sparqlee/blob/master/docs/type-scheme.svg

Tpt: Thank you so much for the review and the suggestion. I just pushed an other attempt 8e9e86c07e53f08623b2e02987dd81ea9599bb40 using the `Literal` `mainSparqlType` field. This makes easy to type check the code and see if it's exhaustive. This way if a new `MainSparqlType` is added the type checker will require to update the ordering code to support it.

The internal sparqlee types are more fine grained and I believe don't need this level of details for the ordering imho. Using directly the `mainSparqlType` also allows to make use of the `typeDiscoveryCallback` evaluation done during literal transformation.

Thank you again for your suggestion. I hope my new proposal is moving things forward.

jitsedesmet: Oh I see now, the stronger/ experimental type system makes sure you still have the correct `mainSparlType` even if the type has to be discovered. The type handling seems good now. As for implementation, that's for Ruben to decide. :slightly_smiling_face:

rubensworks (approved):
Excellent work @Tpt, this looks perfect to me!

PR #140, opened 2022-05-07
Ordering: Adds support of named and blank node ordering

Tpt:
Applies the blank node < named node < literal order from the SPARQL specification.

Restricts the "orderTypes" method argument type to `NamedNode | BlankNode | Literal | undefined` to match which types are actually supported by the ordering operation.

Second step to solve https://github.com/comunica/comunica/issues/892


lib/util/Ordering.ts
```diff
@@ -5,51 +5,82 @@ import type { MainSparqlType } from './Consts';
...
+  if (termA.equals(termB)) {
+    return 0;
+  }
+  return isTermLowerThan(termA, termB, typeDiscoveryCallback, typeCache, enableExtendedXSDTypes) === isAscending ?
+    -1 :
+    1;
 }
 
-function isLowerThan(litA: RDF.Term, litB: RDF.Term,
+function isTermLowerThan(litA: RDF.Term, litB: RDF.Term,
```
rubensworks, line 31:
```suggestion
function isTermLowerThan(termA: RDF.Term, termB: RDF.Term,
```

rubensworks (changes requested):
Looks perfect to me!
Just one minor nit.

PR #142, opened 2022-05-18
Drops regular operator infinite arity support

Tpt:
Not usable in practice because of implementation restrictions

### DIFF
--- a/lib/transformers/AlgebraTransformer.ts
+++ b/lib/transformers/AlgebraTransformer.ts
@@ -124,11 +124,6 @@ export class AlgebraTransformer extends TermTransformer implements IAlgebraTrans
   }
 
   private static hasCorrectArity(args: E.Expression[], arity: number | number[]): boolean {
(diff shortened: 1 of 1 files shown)


rubensworks:
If I understand correctly, this check is in place for varargs expressions such as `CONCAT`.
Do you think this check is invalid?

Tpt:
> If I understand correctly, this check is in place for varargs expressions such as CONCAT.
> Do you think this check is invalid?

The check is valid but unused because `CONCAT` is currently implemented a a "special" operator because regular function evaluation utilities do not support varargs at the moment.

To cover this alternative properly with a test I would need either to do a huge refactoring to add proper dependency injection or hack on the regular function map to add a fake function only for the test and remove it afterward. I prefered to completely drop this piece of code in the assumption that no code is better than dead code.

rubensworks:
Ah I see. Definitely makes sense in that case!

PR #143, opened 2022-05-18
Support both upper and lower case operators

### LINKED ISSUE #21: Test support for uppercase and lowercase function names
We should test how our implementation holds up against function names in the algebra that have a different case. I assume it will break hard. Therefore: let's test and potentially fix.

Tpt:
Closes #21

### DIFF
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -821,8 +821,8 @@ export const definitions: Record<C.RegularOperator, IOverloadedDefinition> = {
   // https://www.w3.org/TR/sparql11-query/#OperatorMapping
   // --------------------------------------------------------------------------
   '!': not,
-  UPLUS: unaryPlus,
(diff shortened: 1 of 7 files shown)


rubensworks (commented):
Would it make sense to also add one (or more?) tests for specific operators testing both uppercase and lowercase invocations?
Such as here: https://github.com/comunica/sparqlee/blob/master/test/integration/functions/op.bnode.test.ts

We could also adapt the `runTestTable` util to execute both variants, but that may be overkill.

Tpt:
> Would it make sense to also add one (or more?) tests for specific operators testing both uppercase and lowercase invocations?

Indeed, I will do so.

Tpt:
> > Would it make sense to also add one (or more?) tests for specific operators testing both uppercase and lowercase invocations?
> 
> Indeed, I will do so.

It's now done.

rubensworks (approved):
Excellent!

PR #144, opened 2022-05-21
Improves function tests code coverage

Tpt:
All the lines in the `functions` directory should be covered except for the `uplus` function.

This PR also fixes float/doubles parsing and serialization of NaN and infinities.

It also removes some dead code in the helpers.

### DIFF
--- a/lib/expressions/Term.ts
+++ b/lib/expressions/Term.ts
@@ -194,6 +194,16 @@ export class DoubleLiteral extends NumericLiteral {
   }
 
   protected specificFormatter(val: number): string {
+    if (!Number.isFinite(val)) {
+      if (val > 0) {
+        return 'INF';
+      }
+      if (val < 0) {
+        return '-INF';
+      }
+      return 'NaN';
+    }
+
     const jsExponential = val.toExponential();
     const [ jsMantisse, jsExponent ] = jsExponential.split('e');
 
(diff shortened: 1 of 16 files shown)


Tpt:
About the "uplus" function, it is not emitted by the SPARQL parser (RubenVerborgh/SPARQL.js/issues/152). I see thre options to archive full code coverage:
1. Drop it from sparqlee
2. Emit it from SPARQL.js
3. Add a "hand written" test to cover it.

@RubenVerborgh What do you think about it?

rubensworks:
Perfect! No comments here :-)

> About the "uplus" function, it is not emitted by the SPARQL parser (https://github.com/RubenVerborgh/SPARQL.js/issues/152). I see thre options to archive full code coverage:

If @RubenVerborgh agrees I would go for option 2: Emit it from SPARQL.js
It might also require a small change in SPARQLAlgebra.js, but @joachimvh might know more about this.

joachimvh:
> It might also require a small change in SPARQLAlgebra.js, but @joachimvh might know more about this.

SPARQLAlgebra.js copies the expression functions directly from SPARQL.js so my guess is it would need no changes (except perhaps an update to the SPARQL.js dependency).

RubenVerborgh:
> If @RubenVerborgh agrees I would go for option 2: Emit it from SPARQL.js

Yes please!

PR #145, opened 2022-05-21
Improves code coverage of Evaluators

Tpt:


lib/aggregators/Max.ts
```diff
@@ -7,10 +7,7 @@ interface IExtremeState {
 export class Max extends BaseAggregator<IExtremeState> {
   public init(start: RDF.Term): IExtremeState {
     const { value } = this.extractValue(start);
-    if (start.termType === 'Literal') {
```
Tpt, line 10: `extractValue` already ensures that `start` is a literal.

lib/evaluators/evaluatorHelpers/AsyncRecursiveEvaluator.ts
```diff
@@ -97,7 +97,7 @@ export class AsyncRecursiveEvaluator extends BaseExpressionEvaluator
   // TODO: Remove?
   private async evalAggregate(expr: E.Aggregate, _mapping: RDF.Bindings): Promise<E.Term> {
     if (!this.context.aggregate) {
-      throw new Err.NoExistenceHook();
```
Tpt, line 100: This is not the correct error.

lib/aggregators/Max.ts
rubensworks: There might be some TS magic we can apply here to automatically let `start` be inferred as Literal.
Something like `start a RDF.Iteral` might be appendable to the return type of `extractValue` (but I don't see this as in-scope for this bounty, just optional :-) ).

rubensworks (approved):
🎉

lib/aggregators/Max.ts
rubensworks: Also sounds good to me :-)

PR #148, opened 2022-05-23
Removes BlankNode.nextID

Tpt:
Unused


Tpt:
It's a breaking change so I am not sure it would be welcomed. If it is better, I can easily cover this function with a test.

rubensworks:
AFAICS, it's not being used in Comunica either. So it's fine to remove this.

PR #153, opened 2022-05-25
Upgrades sparqlalgebrajs

Tpt:
Allows to get a new version of the SPARQL parser supporting unary "+" operator

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "rdf-string": "^1.6.0",
     "relative-to-absolute-iri": "^1.0.6",
     "spark-md5": "^3.0.1",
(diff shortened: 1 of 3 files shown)


rubensworks:
To be safe, I guess we also want a strict version increase (can be patch/minor) of sparqlalgebrajs and the dependency on it here.

@joachimvh could you bump the required version of sparql.js, and bump sparqlalgebrajs?

joachimvh:
> @joachimvh could you bump the required version of sparql.js, and bump sparqlalgebrajs?

Released as v4.0.3

Tpt:
@joachimvh Thank you!
@rubensworks I have updated the PR with an updated revision range.

Issue #37, opened 2019-05-20
Test all functions


wschella:
The spec tests are very insufficient for testing all logic (certainly in regards to types) of all functions. For every function we should add a test ourselves.

jitsedesmet:
I think this issue can be closed? Haven't checked manually, but I don't see how we could have got 100% coverage if we don't test all functions.

wschella:
It originally meant better testing (for all functions). The fact that you have full coverage doesn't necessarily means you test all cases between interacting functions. And the spec tests don't cover much. That said, this issue is a bit too vague, but we should be aware there is room for more tests, especially w.r.t. typing logic.

jitsedesmet:
I agree. Better tests could be beneficial. However, since we can not test **everything** and I don't see any benefit in transferring this issue to Comunica, I will close it.

Issue #35, opened 2019-05-20
Test general concepts


wschella:
There are a lot of general concepts (error handling, type promotion, EBV), that are implemented, but at the best loosely tested trough some unrelated tests.

Things we should definitely still test:

- [ ] Error handeling
- [ ] Subtype subsitution
- [ ] Type promotion
- [ ] Handling of nonlexicals
- [ ] Coercing to EBV
- [ ] Exists / aggregate hook

jitsedesmet:
My report:

For error handling, we should have the error table. Even though we could write more tests, our test suite allows us to easily add them, and there are already a few locations where we use them. I suggest we ignore this for now.

For Subtype substitution, I will implement some more tests. 

Type promotion has some test introduced in https://github.com/comunica/sparqlee/pull/103/files#diff-4db5dd024582568f1ac1245e2af228c5611a35cca8d33e741db5ac081ddd2c28

Handling of non-Lexical forms has some minimal tests. (for example, in addition)

Will add some tests regarding Coercing to EBV.

Exists has a test https://github.com/jitsedesmet/comunica/blob/master/packages/expression-evaluator/test/integration/misc/Exists-test.ts
I don't think we have a test for the aggregate hook. The creation of a Comunica engine test might bé the best way to test them. (@rubensworks now that sparqlee is a package within Comunica this should suffice to close this issue?)

rubensworks:
> I don't think we have a test for the aggregate hook. The creation of a Comunica engine test might bé the best way to test them. 

Sounds good!

PR #157, opened 2022-05-31
100% code coverage

### LINKED ISSUE #38: Achieve full test coverage

Tpt:
Covers the remaining pieces of code and enforces 100% code coverage

Closes #38

### DIFF
--- a/lib/transformers/TermTransformer.ts
+++ b/lib/transformers/TermTransformer.ts
@@ -63,8 +63,6 @@ export class TermTransformer implements ITermTransformer {
     }
     const dataType = lit.datatype.value;
     switch (dataType) {
-      case null:
-      case undefined:
       case '': {
         return lit.language ?
           new E.LangStringLiteral(lit.value, lit.language) :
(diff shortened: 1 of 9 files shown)


rubensworks:
Excellent work @Tpt! 🎉 

Do you think there are any other test-related issues that can be closed now? (not saying there should be)

Tpt:
> Excellent work @Tpt!

Thank you!

> I guess this completes the bounty?

Yes, if you think so. 

> Do you think there are any other test-related issues that can be closed now? (not saying there should be)

I believe #35 is at least partially fulfilled (maybe some small parts are still missing like "Subtype substitution" that won't be highlighted by code coverage).
About #37, all functions are now covered by either Sparqlee hand written tests or the port of the official test suite but I don't think all behaviors of all all functions are tested (there are shared utility functions so if each piece of the utility functions is covered by one test for one function is enough to get 100% test coverage). But testing all behaviors of all functions would be quite a lot of work (it means e.g. test the full EBV behavior for each logical function...).

PR #162, opened 2022-11-29
Remove old type system

jitsedesmet:
This PR aims to remove the old type system alltogether. I discovered an optimization that made the experimental type system at least as fast as the old one. Since the new one is stronger and easier to maintain, we will use that one as default. It was also a pain to make sure the code did the correct thing in both type systems.

It will be noted that one order test broke and I just switched them. IRC the only importance in that scenario was that the order had some concistency.

The PR deprecates the `overloadCache` and  `enableExtendedXsdTypes` config options.

PS: I'm sorry my first PR in a while is this big hehe.


jitsedesmet:
Both the tests and spec test vahe been run on the communica repo and they pass.

benchmarks/integerAddition.ts
```diff
@@ -6,7 +6,6 @@ import type * as RDF from '@rdfjs/types';
 import type { Event } from 'benchmark';
 import { Suite } from 'benchmark';
 import * as Benchmark from 'benchmark';
-import * as LRUCache from 'lru-cache';
```
rubensworks, line 9: Can it be removed from our package.json as well?

benchmarks/integerAddition.ts
```diff
@@ -21,47 +20,7 @@ function integerTerm(int: number): RDF.Term {
...
-        [ DF.variable('a'), integerTerm(fst) ],
-        [ DF.variable('b'), integerTerm(snd) ],
-        [ DF.variable('c'), integerTerm(fst + snd) ],
-      ]));
-    }
-  }
-});
-
-const nonExperimental = new Benchmark('bench addition non experimental', () => {
+const cache = new Benchmark('bench addition', () => {
```
rubensworks, line 23:
```suggestion
const benchmark = new Benchmark('bench addition', () => {
```

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
```diff
@@ -9,14 +10,14 @@ import type { SuperTypeCallback, TypeCache, ISuperTypeProvider } from '../../uti
 export interface ISharedContext {
   now?: Date;
   baseIRI?: string;
-  overloadCache?: OverLoadCache;
+  /**
+   * @deprecated Deprecated since version ... provided value will be ignored.
```
rubensworks, line 14: Is that correct? Because I see you're still passing values to it?
But the values are just `{}`, which are not applicable for `LRUCache`.

rubensworks, line 14:
```suggestion
   * @deprecated This value is unused, will be removed in the next major version.
```

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
```diff
@@ -9,14 +10,14 @@ import type { SuperTypeCallback, TypeCache, ISuperTypeProvider } from '../../uti
...
+   */
+  overloadCache?: LRUCache<string, any>;
   typeCache?: TypeCache;
   getSuperType?: SuperTypeCallback;
   /**
-   * This feature is opt in. It activates the use of a new 'experimental' type system.
-   * This system is needed when using typeCache, overloadCache or getSuperType.
-   * The system is more powerful and reliable. In most cases however the old system works perfectly.
-   * Using this experimental system makes sparqlee a bit slower but more reliable using type promotion for example.
+   * @deprecated Deprecated since version ... . The type 'experimental' system is always used.
```
rubensworks, line 20:
```suggestion
   * @deprecated This value is unused (extended XSD types is always enabled), will be removed in the next major version.
```

lib/util/Ordering.ts
```diff
@@ -2,10 +2,14 @@ import type * as RDF from '@rdfjs/types';
...
 import { TermTransformer } from '../transformers/TermTransformer';
-import type { MainSparqlType } from './Consts';
-import type { ISuperTypeProvider, SuperTypeCallback, TypeCache } from './TypeHandling';
+import { TypeAlias, TypeURL } from './Consts';
+import type { ISuperTypeProvider, SuperTypeCallback, TypeCache, GeneralSuperTypeDict } from './TypeHandling';
+import { getSuperTypeDict } from './TypeHandling';
 
 // Determine the relative numerical order of the two given terms.
+/**
+ * @param enableExtendedXSDTypes deprecated since version ... . System will behave like when this was true.
```
rubensworks, line 11:
```suggestion
 * @param enableExtendedXSDTypes System will behave like when this was true. @deprecated
```

rubensworks (changes requested):
Looks good! Just some minor nits.

I really like big code removals like these :-)

benchmarks/integerAddition.ts
jitsedesmet: Good catch. Seems like it, yes!

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
jitsedesmet: This should be correct, yes. So the `ISharedContext` is what the user will provide an engine. And an `ICompleteSharedContext` is what the engine will use. I kept the cache in `ICompleteSharedContext` because that way, each engine has it's own cache which is nice for testing.

I hope this answers your question?

lib/util/Ordering.ts
jitsedesmet: Seems like it doesn't technically do anything to add the `@deprecated` but at least in this case it is indicated (be it in string format).

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
rubensworks: Ok, but maybe we should make it of type `any` then, so that we can remove the typings dependency on lru-cache.

benchmarks/integerAddition.ts
jitsedesmet: Woeps, completly forgot that the openWorld type discoverer also uses the LRUCache. I'll add it back in.

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
jitsedesmet: Seeing how we need the LRUCache anyway, do we keep the `LRUCache` type here or the `any` type?

rubensworks: Sure, we can keep it then.

lib/evaluators/AsyncEvaluator.ts
```diff
@@ -26,7 +26,7 @@ export class AsyncEvaluator {
     return {
       now: context.now || new Date(Date.now()),
       baseIRI: context.baseIRI || undefined,
-      overloadCache: context.overloadCache || new LRUCache(),
+      overloadCache: {},
```
rubensworks, line 29: I just realized an important thing.
I think we _do need_ to expose this field to the end-user, because we want this cache to be used Comunica-wide, across all actors.

So in Comunica, we would store this cache in the context, and pass this along to all actors that invoke sparqlee.

jitsedesmet: That's a good point! How would you suggest solving this? (thinking about the change of types and how to avoid a major update)

rubensworks: Either a different key, or a union type?

jitsedesmet: I'd think a different key would be best? `fuctionArgumentsCache`?

rubensworks: Sure, but maybe without the typo :P

jitsedesmet: Not shure If we should change the name of the OverloadCache too. I feel like this name is most descriptive so internally it might be better to use this name still?

jitsedesmet: Is there any place in comunica where multiple engines are created for the same data? If not keeping the cache might cause more harm to the communica code than that it is worth.

rubensworks: > Not shure If we should change the name of the OverloadCache too. I feel like this name is most descriptive so internally it might be better to use this name still?

Sure, that's fine.

> Is there any place in comunica where multiple evaluators are created for the same data?

I see you already found it in the other PR.

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
```diff
@@ -1,32 +1,35 @@
...
   getSuperType?: SuperTypeCallback;
   /**
-   * This feature is opt in. It activates the use of a new 'experimental' type system.
-   * This system is needed when using typeCache, overloadCache or getSuperType.
-   * The system is more powerful and reliable. In most cases however the old system works perfectly.
-   * Using this experimental system makes sparqlee a bit slower but more reliable using type promotion for example.
+   * @deprecated This value is unused (extended XSD types is always enabled), will be removed in the next major version.
    */
   enableExtendedXsdTypes?: boolean;
+  // This name is chosen because overloadCache was already taken by previous version.
```
rubensworks, line 23: The downside of this is that we require downstream users to modify things twice, which is not so nice.
As alternative, what about just renaming `OverLoadCache` to `FunctionArgumentsCache`?

jitsedesmet: That is better indeed! Otherwise, things might get confusing :)

rubensworks (approved):
Looks perfect, great work!

rubensworks:
@jitsedesmet Before I release this, can you double-check on your end that all tests (+spec and integration) still pass in Comunica with this version?

PR #163, opened 2023-01-03
Additional date support - sparql 1.2

jitsedesmet:
Sparql 1.2 suggests additional date support on [their GitHub spec page](https://github.com/w3c/sparql-12/blob/main/SEP/SEP-0002/sep-0002.md). This PR implements the suggested features.
In addition to implementing these features, this PR should also close 2 issues of the Comunica repo, mainly:
https://github.com/comunica/comunica/issues/1065 and https://github.com/comunica/comunica/issues/1098 .

The PR is expected to grow pretty large, because sparqlee previously used the fact that Sparql didn't require the creation or manipulation of dates. Implementing the functionality requires rewriting how sparqlee looks at dates.

The aim has been to use JS Date as much as possible, every feedback is welcome in regard to how this was done.


jitsedesmet:
@wschella During implementation of this PR I needed to create some more regular function (you don't need to look at the PR, don't worry). 
While implementing this, I started wondering whether our handling of `InvalidLiteral` is still optimal.
If I remember correctly, this type is created when the `TermTransformer` gets an invalid lexical notation instead of throwing an error because there are functions that should still work on them.  
Now we handle the error trowing in our helper like: https://github.com/comunica/sparqlee/blob/master/lib/functions/Helpers.ts#L213-L214

It seems like some functions now just don't have this kind of handling since you need to use one of the helper creators.
I suggest the `.collect()` function (optionally) adds these overloads, so they don't need to be added manually.

 I wanted to get this by you just to make sure I still understand these types correctly (I remember I was confused about this in the past).
 
 Anyways, Thanks :D

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,195 @@
...
+    'hours',
+    'minutes',
+    'seconds',
+    'day',
+  ].includes(key))) {
+    throw new Error('nono');
+  }
+  return res;
+}
+
```
jitsedesmet, line 179: I am wondering if I could do this in a better way. @rubensworks Do you have a suggestion? (Don't mind the error messages)

rubensworks: This might be a bit more compact, and slightly more performant: 

`if([ 'year', 'month' ].some(key => !(key in res)))`

But this assumes that keys in `res` always have a defined value.

(if this approach works, I would suggest moving this constant array outside of this function as an actual constant, to avoid re-recreating it for every call.

jitsedesmet: They do not always have a defined value. My biggest issue is actually that both `year` and `month` in the first function are the 2 keys in `IYearMonthDurationRepresentation`.  So this code gets some typescript keys into JS runtime, but the problem is ofcource that the compiler will not catch the renaming of the `year` field into `years` for example. So basically I would want to get a list of all the keys for the `IYearMonthDurationRepresentation` interface. 
As in this issue: https://stackoverflow.com/questions/43909566/get-keys-of-a-typescript-interface-as-array-of-strings

rubensworks: Yeah, that's not possible indeed, as TS typings are not available at runtime.

But if `undefined` is possible, then the following may work:

`if([ 'year', 'month' ].some(key => !res[key]))`

jitsedesmet: Alright, that's unfortunate. I'll make sure to add some comments than :)
Your code fragment also looks better :) Thank you!

wschella:
I think the reason handling of invalid lexical forms is done semi-manually
now is because the spec says exactly these should still work on invalid
lexical forms and the others should explicitly not.

If current behaviour is not according to the new spec, or you find a
simpler way to keep the same behaviour, feel free to rewrite of course :)

Wout

jitsedesmet:
Alright, thank you, @wschella ! One more thing, you say:
> the spec says exactly these should still work on invalid lexical forms and the others should explicitly not.

What if nothing is said? In this case it is right to assume the function should not work right? In other words, the following error should be thrown: https://github.com/comunica/sparqlee/blob/master/lib/functions/Helpers.ts#L250L254

wschella:
I think so, the default for functions invoked with wrong types is to error
according to https://www.w3.org/TR/sparql11-query/#evaluation.

Wout

lib/transformers/TermTransformer.ts
```diff
@@ -67,13 +75,27 @@ export class TermTransformer implements ITermTransformer {
...
       return new E.LangStringLiteral(lit.value, lit.language);
     }
-    if (DT.XSD_DATE_TIME in superTypeDict) {
+    if (TypeURL.XSD_YEAR_MONTH_DURATION in superTypeDict) {
+      return new E.YearMonthDurationLiteral(yearMonthDurationParser(lit.value), lit.value, dataType);
+    }
+    if (TypeURL.XSD_DAY_TIME_DURATION in superTypeDict) {
+      return new E.DayTimeDurationLiteral(dayTimeDurationParser(lit.value), lit.value, dataType);
+    }
+    // TODO: maybe the special notations need to be checked so we don't Years in a DayTimeDuration? ...
```
jitsedesmet, line 91: > maybe the special notations need to be checked so we don't have a Years field in a DayTimeDuration?

I think this is done? But yes. This should be checked since it would be an invalid literal Form!

lib/util/Consts.ts
```diff
@@ -24,6 +24,13 @@ export enum TypeURL {
   XSD_DATE_TIME_STAMP = 'http://www.w3.org/2001/XMLSchema#dateTimeStamp',
   XSD_DATE = 'http://www.w3.org/2001/XMLSchema#date',
 
+  XSD_G_MONTH = 'http://www.w3.org/2001/XMLSchema#gMonth',
+  XSD_G_MONTHDAY = 'http://www.w3.org/2001/XMLSchema#gMonthDay',
+  XSD_G_YEAR = 'http://www.w3.org/2001/XMLSchema#gYear',
+  XSD_G_YEAR_MONTH = 'http://www.w3.org/2001/XMLSchema#gYearMonth',
+  XSD_TIME = 'http://www.w3.org/2001/XMLSchema#time',
+  XSD_G_DAY = 'http://www.w3.org/2001/XMLSchema#gDay',
```
jitsedesmet, line 32: I added these types here just because they are in the graph of supported types. We don't do anything with them, unfortunately. 

It seems like the spec is not shure what to do with them as they are not listed in: https://github.com/w3c/sparql-12/blob/main/SEP/SEP-0002/sep-0002.md
And are mentioned in a TODO question in the test files: https://github.com/kasei/sparql-12/tree/xsd_datetime_duration/tests/xsd_functions

@rubensworks would it be interesting to explicitly ask/ reunite the discussion on the github spec or do I just leave this unsupported?

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+// Parsers:
+export function dateTimeParser(dateTimeStr: string, errorCreator?: () => Error): IDateTimeRepresentation {
+  // https://www.w3.org/TR/xmlschema-2/#dateTime
+  const [ date, time ] = dateTimeStr.split('T');
+  return { ...dateParser(date, errorCreator), ...internalTimeParser(time, errorCreator) };
+}
+
+export function dateTimeSerializer(date: IDateTimeRepresentation): string {
+  // Need to extract the date because we don't want the timeZone part here
```
jitsedesmet, line 22:
```suggestion
  // Extraction is needed because the date serializer can not add timezone yet.
```

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+    day: duration[3] ? sign * Number(duration[3]) : undefined,
+    hours: duration[4] ? sign * Number(duration[4]) : undefined,
+    minutes: duration[5] ? sign * Number(duration[5]) : undefined,
+    seconds: duration[6] ? sign * Number(duration[6]) : undefined,
+  };
+}
+
+export function yearMonthDurationParser(durationStr: string): Partial<IYearMonthDurationRepresentation> {
+  const res = durationParser(durationStr);
+  if (Object.entries(res).some(([ key, value ]) => value !== undefined && ![ 'year', 'month' ].includes(key))) {
```
jitsedesmet, line 156: Make this list a const and look at the comment provided by Ruben!

lib/util/InternalRepresentations.ts
```diff
@@ -0,0 +1,181 @@
+/**
+ * We redefine some types here since the JS Date implementation is not sufficient.
+ */
```
jitsedesmet, line 3: Can go.

lib/util/InternalRepresentations.ts
```diff
@@ -0,0 +1,181 @@
...
+  return new Date(
+    localTime.getTime() + (
+      -1 * (date.zoneHours === undefined ? defaultTimezone.zoneHours : date.zoneHours) * 60 -
+      (date.zoneMinutes === undefined ? defaultTimezone.zoneMinutes : date.zoneMinutes) -
+      localTime.getTimezoneOffset()
+    ) * 60 * 1_000,
+  );
+}
+
+export function ensureYearMonthDuration(dur: Partial<IDurationRepresentation>):
```
jitsedesmet, line 149: Bad naming. This function ensures your object ONLY contains this. The name might suggest it checks whether it contains this.

lib/util/specAlgos.ts
```diff
@@ -0,0 +1,108 @@
+import type {
+  IDurationRepresentation,
+  IDateTimeRepresentation, ITimeZoneRepresentation,
+} from './InternalRepresentations';
+import { toUTCDate } from './InternalRepresentations';
+
```
jitsedesmet, line 6: Add documentation as to what this file does.

jitsedesmet: + Might be interesting to rename/ remove the used functions and where possible make more speedy.

test/integration/evaluators/RecursiveEvaluator.test.ts
```diff
@@ -21,6 +21,7 @@ describe('recursive evaluators', () => {
         cache: new LRUCache(),
         discoverer: _ => 'term',
       },
+      defaultTimeZone: { zoneMinutes: 0, zoneHours: 0 },
```
jitsedesmet, line 24: This comes back a lot. Make this a const.

test/spec/additional-data-spec/date_compare.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { bool, dateNotation } from '../../util/Aliases';
+import { Notation } from '../../util/TestTable';
+import { runTestTable } from '../../util/utils';
+
+describe('compare date', () => {
+  describe('respect the op:date-equal xpath-functions spec', () => {
+    // Originates from: https://www.w3.org/TR/xpath-functions/#func-date-equal
```
jitsedesmet, line 7: Spec tests might not be the best dir for this file.

test/spec/additional-data-spec/time_compare.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { bool, timeNotation } from '../../util/Aliases';
+import { Notation } from '../../util/TestTable';
+import { runTestTable } from '../../util/utils';
+
+describe('compare date', () => {
+  describe('respect the op:time-equal xpath-functions spec', () => {
+    // Originates from: https://www.w3.org/TR/xpath-functions/#func-time-equal
```
jitsedesmet, line 7: Might not be the best dir for this file.

jitsedesmet (commented):
Just read through the code to find what still needs to happen in this PR.

README.md
```diff
@@ -353,6 +353,17 @@ which you can provide as `baseIRI: string` to the config.
 | dT             | ✓ | ✓ | ? |   |   |
 | bool           | ✓ | ✓ | ? |   |   |
 
+### Sparql 1.2
+
+Sparqlee looks forward and already implements some Sparql 1.2 specification functions.
+
+Currently, this is restricted to the [extended date](https://github.com/w3c/sparql-12/blob/main/SEP/SEP-0002/sep-0002.md) functionality.
+Please note that the new build in ADJUST function has not been implemented due to dependencies.
```
rubensworks, line 361:
```suggestion
Please note that the new built-in ADJUST function has not been implemented due to dependencies.
```

README.md
```diff
@@ -353,6 +353,17 @@ which you can provide as `baseIRI: string` to the config.
 | dT             | ✓ | ✓ | ? |   |   |
 | bool           | ✓ | ✓ | ? |   |   |
 
+### Sparql 1.2
+
+Sparqlee looks forward and already implements some Sparql 1.2 specification functions.
```
rubensworks, line 358:
```suggestion
Sparqlee looks forward and already implements some SPARQL 1.2 specification functions.
```

README.md
```diff
@@ -353,6 +353,17 @@ which you can provide as `baseIRI: string` to the config.
 | dT             | ✓ | ✓ | ? |   |   |
 | bool           | ✓ | ✓ | ? |   |   |
 
+### Sparql 1.2
```
rubensworks, line 356:
```suggestion
### SPARQL 1.2
```

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
```diff
@@ -21,13 +22,16 @@ export interface ISharedContext {
    */
   enableExtendedXsdTypes?: boolean;
   functionArgumentsCache?: FunctionArgumentsCache;
+  // https://www.w3.org/TR/xpath-functions/#func-implicit-timezone (Redefined to reduce coupling to internal types)
```
rubensworks, line 25: > (Redefined to reduce coupling to internal types)

I think it's fine to just use `ITimeZoneRepresentation`.
As long as we `export` this type form sparqlee, others can use it.

lib/functions/NamedFunctions.ts
```diff
@@ -154,6 +162,65 @@ const xsdToBoolean = {
...
+    .onStringly1(() => (val: Term) =>
+      new E.DayTimeDurationLiteral(dayTimeDurationParser(val.str()), val.str()))
+    .collect(),
+};
+
+const xsdToYearMonthDuration = {
+  arity: 1,
+  overloads: declare(TypeURL.XSD_YEAR_MONTH_DURATION)
+    // https://www.w3.org/TR/xpath-functions/#casting-to-durations
+    // TODO: make a test for the case of dayTime to YearMonth...
```
rubensworks, line 215: TODO

lib/functions/OverloadTree.ts
```diff
@@ -170,24 +168,11 @@ export class OverloadTree {
...
-    }
-    if (argumentType === TypeURL.XSD_FLOAT) {
-      this.addPromotedOverload(TypeURL.XSD_DECIMAL, func, arg =>
-        float((<E.NumericLiteral>arg).typedValue), _argumentTypes, promotionCount);
-    }
+
+    typePromotion[argumentType]?.forEach(ret =>
+      this.addPromotedOverload(
+        ret.typeToPromote, func, ret.conversionFunction, _argumentTypes, promotionCount,
+      ));
```
rubensworks, line 175: Why was this change needed exactly?

lib/functions/RegularFunctions.ts
```diff
@@ -641,17 +783,15 @@ const rand = {
...
-
 /**
  * https://www.w3.org/TR/sparql11-query/#func-now
  */
 const now = {
   arity: 0,
   overloads: declare(C.RegularOperator.NOW).set([], (sharedContext: ICompleteSharedContext) => () =>
-    new E.DateTimeLiteral(sharedContext.now, sharedContext.now.toISOString())).collect(),
+    new E.DateTimeLiteral(toDateTimeRepresentation(
+      { date: sharedContext.now, timeZone: sharedContext.defaultTimeZone },
```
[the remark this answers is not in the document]
rubensworks: Does it really change the behaviour in a breaking way?
I guess it just enhances it with the default timezone, right?

lib/transformers/TermTransformer.ts
```diff
@@ -67,13 +75,27 @@ export class TermTransformer implements ITermTransformer {
...
+    if (TypeURL.XSD_DURATION in superTypeDict) {
+      // TODO: All of these new parsers can fail, a failure should result in a NonLexicalLiteral!
```
rubensworks, line 93: Also a TODO here.

lib/util/Consts.ts
rubensworks: > @rubensworks would it be interesting to explicitly ask/ reunite the discussion on the github spec or do I just leave this unsupported?

Let's leave it unsupported for this PR.
But if you can think of a reasonable implementation of them, feel free to implement them in a next PR.
Based on that implementation, we can contribute support for them in the SPARQL-12 SEP.

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
+import type {
+  IDateRepresentation, IDateTimeRepresentation, IDayTimeDurationRepresentation,
+  IDurationRepresentation,
+  ITimeRepresentation,
+  ITimeZoneRepresentation, IYearMonthDurationRepresentation,
+} from './InternalRepresentations';
+import { extractDayTimeDur } from './InternalRepresentations';
+import { maximumDayInMonthFor } from './specAlgos';
+
+function numSer(num: number, min = 2): string {
```
rubensworks, line 10: Not sure what `numSer` means. Could we use a more descriptive name?

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+  return num.toLocaleString(undefined, { minimumIntegerDigits: min, useGrouping: false });
+}
+
+// Parsers:
+export function dateTimeParser(dateTimeStr: string, errorCreator?: () => Error): IDateTimeRepresentation {
```
rubensworks, line 15: `dateTimeParser` (and similar below) are a bit confusing, as it sounds like this is an object, while it is just a function.
Something like `dateTimeParse` or `parseDateTime` might be better names.

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+  const date = dateStrings.slice(0, -1).map(str => Number(str));
+
+  const res = {
+    year: date[0] * date[1],
+    month: date[2],
+    day: date[3],
+    ...timeZoneParser(dateStrings[4]),
+  };
+  if (!(1 <= res.month && res.month <= 12) || !(1 <= res.day && res.day <= maximumDayInMonthFor(res.year, res.month))) {
+    throw new Error('nono');
```
rubensworks, line 70: Better error message? :-)

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+  const extraction = /(Z|([+-]\d\d:\d\d))?$/u.exec(zoneContained);
+  // It is safe to cast here because the empty string can always match.
+  return extraction![0];
+}
+
+export function dateSerializer(date: IDateRepresentation): string {
+  return `${numSer(date.year, 4)}-${numSer(date.month)}-${numSer(date.day)}${timeZoneSerializer(date)}`;
+}
+
+function internalTimeParser(timeStr: string, errorCreator?: () => Error): ITimeRepresentation {
```
rubensworks, line 85: Is it really "internal" or is it just a time-specific parser?

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+  const res = {
+    hours: time[0],
+    minutes: time[1],
+    seconds: time[2],
+    ...timeZoneParser(timeStrings[3]),
+  };
+
+  if (res.seconds >= 60 || res.minutes >= 60 || res.hours > 24 ||
+    (res.hours === 24 && (res.minutes !== 0 || res.seconds !== 0))) {
+    throw new Error('nono');
```
rubensworks, line 103: Better error message.

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+  if (formattedDayDur === dayNotation) {
+    throw new Error('nono');
+  }
+
+  const durationStrings = formattedDayDur.split('!');
+  if (timeNotation !== undefined) {
+    const formattedTimeDur = timeNotation.replace(/^(\d+H)?(\d+M)?(\d+(\.\d+)?S)?$/gu, '$1!$2!$3');
+
+    if (timeNotation === '' || timeNotation === formattedTimeDur) {
+      throw new Error('nono');
```
rubensworks, line 134: TODO

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+    }
+    durationStrings.push(...formattedTimeDur.split('!'));
+  }
+  const duration = durationStrings.map(str => str.slice(0, -1));
+  if (!duration.slice(1).some(item => item)) {
+    throw new Error('nono');
```
rubensworks, line 140: TODO

test/spec/additional-data-spec/adjust_date-01.spec.ts
```diff
@@ -0,0 +1,62 @@
+/* eslint max-len: 0 */
```
rubensworks, line 1: Instead of `additional-data-spec`, perhaps we can use `sparql-12/sep-0002`.

rubensworks (changes requested):
This is really amazing work @jitsedesmet, I'm impressed!

It required a lot more code than I initially expected, so I definitely underestimated this task.

lib/util/InternalRepresentations.ts
```diff
@@ -0,0 +1,181 @@
...
+export function toUTCDate(date: Partial<IDateTimeRepresentation>,
+  defaultTimezone: ITimeZoneRepresentation): Date {
+  const localTime = toJSDate(defaultedDateTimeRepresentation(date));
+  // This date has been constructed in machine local time, now we alter it to become UTC and convert to correct timezone
+  return new Date(
+    localTime.getTime() + (
+      -1 * (date.zoneHours === undefined ? defaultTimezone.zoneHours : date.zoneHours) * 60 -
+      (date.zoneMinutes === undefined ? defaultTimezone.zoneMinutes : date.zoneMinutes) -
+      localTime.getTimezoneOffset()
+    ) * 60 * 1_000,
```
wschella, line 145: This could use some comments and/or some clear code. It's a lot at once.

lib/util/DateTimeHelpers.ts
```diff
@@ -0,0 +1,191 @@
...
+  return `${negative ? '-' : ''}P${dur.year ? `${Math.abs(dur.year)}Y` : ''}${dur.month ?
+    `${Math.abs(dur.month)}M` :
+    ''}${dur.day ?
+    `${Math.abs(dur.day)}D` :
+    ''}${hasTimeField ?
+    `T${dur.hours ? `${Math.abs(dur.hours)}H` : ''}${dur.minutes ?
+      `${Math.abs(dur.minutes)}M` :
+      ''}${dur.seconds ? `${Math.abs(dur.seconds)}S` : ''}` :
+    ''}`;
+}
```
wschella, line 191: This would need some comments and/or more clear code. A lot is going on at once. An example input/output might also be very nice.

lib/util/specAlgos.ts
```diff
@@ -0,0 +1,108 @@
+import type {
```
wschella, line 1: I assume file needs to be capitalized.

test/integration/evaluators/RecursiveEvaluator.test.ts
wschella: Or make it an optional argument in the evaluator, and use this by default.

lib/expressions/DateLiterals.ts
```diff
@@ -0,0 +1,65 @@
+import { TypeURL } from '../util/Consts';
```
wschella, line 1: DateTimeLiterals would a name more consistent with the other naming.
Additionally, I don't see why they go into a separate file. If put separately, I suggest to put all literals separately instead, either in a single `Literals.ts` or with one file for each type. Of course, if Literals are separate, so should all the other Term subtypes to keep things consistent and simple.

test/spec/additional-data-spec/adjust_date-01.spec.ts
wschella: I agree!

test/util/Aliases.ts
```diff
@@ -89,11 +89,51 @@ export function double(value: string): string {
...
+export function dateTimeNotation(value: string): string {
   return compactTermString(value, 'xsd:dateTime');
 }
 
-function compactTermString(value: string, dataType: string): string {
+/**
+ * ''02:12:59'' => "'02:12:59'"^^xsd:time
+ * @param value string (representing a date)
+ */
+export function timeNotation(value: string): string {
```
wschella, line 100: The longer function names seem too vague and ambiguous for their length. It's also not consistent with the other short ones like int and decimal. Is there a reason these ones have 'Notation' added?
How about these: `xsd_time(val)`, `xsdTime(val)`,  `typedTime(val)`, `timeTyped(val)`, `addTimeType(val)` or `withTimeType(val)`?

wschella (approved):
Nice work!

One comment (and a few in the files):
I don't see why DateTimeHelper and InternalRepresenation are separate files. They are tightly coupled, InternalRepresentation only deals with Dates and Times, and also contains many helper functions. It seems cleaner to put them together in a single util/Time.ts or util/DateTime.ts. This would mimick behaviour of most languages' standard libraries. Alternatively and or jointly, parsing serialising logic goes into util/Parsing.ts.

lib/functions/RegularFunctions.ts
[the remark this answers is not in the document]
jitsedesmet: Right now, when a date is given to the `now` context item, the timezone is always there since a date created in JS will have your timezone. When we use this code, the timeZone will be overwritten by the implicit time zone UTC by default). This behavior could potentially break code for our users. 

The issue could be solved if we don't default to UTC when a dateTime is not provided but just use the undefined in our context. This function could then check if it was provided and based on that use the users time zone.

The other functions could still default to UTC (which is in those function definitely what we want for stability reasons). I don't know what would be best here since having special treatments for one function also isn't pretty?

lib/util/Consts.ts
jitsedesmet: They can all have reasonable implementation. The groundwork has been laid out. And functions like checking equality, gt, lt, .. are all nice to have.

lib/functions/RegularFunctions.ts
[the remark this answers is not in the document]
rubensworks: Perhaps defaulting the timezone to UTC is the problem.
Maybe we just always want to default this to `now`, which is either user-provided, or system-default.

test/integration/evaluators/RecursiveEvaluator.test.ts
jitsedesmet: Seems like the `SyncRecursiveEvaluator` expects a complete context though :'(

lib/functions/OverloadTree.ts
jitsedesmet: Ah yes. It was not. When starting this PR we talked about using some kind of type promotion. That's when I wrote this. The PR shifted to a whole new thing since then. I still think this new way is the better way. TypePromotion is a type handling problem. However, it does not being in this PR. Should I make a separate one for this?

lib/util/DateTimeHelpers.ts
jitsedesmet: Number serializer :)

lib/util/DateTimeHelpers.ts
jitsedesmet: It's a time parser that parses time with accordance to the spec. XSD:Time however still requires a `%24` operation on the hours so in that sense it is not a time parser.

test/util/Aliases.ts
jitsedesmet: I like the timeTyped option. :D

lib/expressions/DateLiterals.ts
jitsedesmet: I think it might be worth it to make a `Literals.ts` file. I moved this to their own file because term was about to go over 350 lines.

jitsedesmet: Hmm, it doesn't seem like that would help much so I'll just keep the big file :)

lib/util/DateTimeHelpers.ts
jitsedesmet: I don't think I can provide cleaner code. I't just some string formatting I'm afraid. I can however give an example input/ output.

lib/functions/OverloadTree.ts
wschella: Type promotion is a finicky but important thing. A separate PR would be nice.

lib/util/DateTimeHelpers.ts
wschella: Just extracting the parts into separate statements would help.
E.g.
```
const sign = negative ? '-' : '';
const year = dur.year ? `${Math.abs(dur.year)}Y` : ''
# ... etc
```
Especially since the linebreaks are in weird places for readability.

lib/functions/RegularFunctions.ts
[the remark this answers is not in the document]
jitsedesmet: It could be seen as the problem, yes.
The issue is you want this to be kind of stable. What this means is that you don't want your evaluation of:
`"2023-03-01T15:00" = "2023-03-01T15:00Z"` to change just by entering a new time zone (and even weirder, day time saving)

lib/functions/OverloadTree.ts
rubensworks: I agree with going for a separate PR on this 👍

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
jitsedesmet: I don't really understand what you mean here. Could you elaborate?

lib/util/DateTimeHelpers.ts
rubensworks: > It's a time parser that parses time with accordance to the spec

Which spec is that? Can we refer to that spec, instead of using `internal`?

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
rubensworks: `  defaultTimeZone?: { hours: number; minutes: number };` -> `  defaultTimeZone?: ITimeZoneRepresentation;`

jitsedesmet: Oh, yes! But for the type export nothing special needs to happen right? It's not like I need to add something to our `index.ts`?

lib/util/DateTimeHelpers.ts
jitsedesmet: This like is referenced in that function: https://www.w3.org/TR/xmlschema-2/#time-lexical-repr
Maybe the name should be `sharedParseTime` since it preforms the parse for multiple both `dateTime` and `time` but the `time` just needs one additional step (`% 24` on the hour field).

lib/evaluators/evaluatorHelpers/BaseExpressionEvaluator.ts
wschella: You might need to, depends on what is already there.

lib/util/DateTimeHelpers.ts
wschella: I think I understand the confusion. I was also wondering why it wasn't called just `timeParser`, but I assume you did it because it's a helper function reused by both `time` and `dateTime` functions. What you really want is to not export this function out of the file, but that's hard, so you just name it something to avoid polluting the namespace.
I think convention for this would be `__parseTime` or `privateParseTime`.

jitsedesmet: Yes, exactly! I changed it to `__parseTime`.

lib/functions/RegularFunctions.ts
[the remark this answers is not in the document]
rubensworks: > What this means is that you don't want your evaluation of:
"2023-03-01T15:00" = "2023-03-01T15:00Z" to change just by entering a new time zone

But that is exactly what we _do_ want to happen when changing the timezone, right?

lib/util/Consts.ts
jitsedesmet: @rubensworks I will just leave this as is (in this PR)?

lib/functions/RegularFunctions.ts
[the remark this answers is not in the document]
rubensworks: That's my interpretation of the spec at least, but if you have a different interpretation, do let me know, because I could be wrong of course :-)

lib/util/Consts.ts
rubensworks: Yep, as-is is fine.

README.md
```diff
@@ -353,6 +353,17 @@ which you can provide as `baseIRI: string` to the config.
 | dT             | ✓ | ✓ | ? |   |   |
 | bool           | ✓ | ✓ | ? |   |   |
 
+### SPARQL 1.2
+
+Sparqlee looks forward and already implements some SPARQL 1.2 specification functions.
+
+Currently, this is restricted to the [extended date](https://github.com/w3c/sparql-12/blob/main/SEP/SEP-0002/sep-0002.md) functionality.
+Please note that the new build-in `ADJUST` function has not been implemented due to package dependencies.
```
rubensworks, line 361:
```suggestion
Please note that the new built-in `ADJUST` function has not been implemented due to package dependencies.
```

README.md
```diff
@@ -353,6 +353,17 @@ which you can provide as `baseIRI: string` to the config.
...
+
+For the extended date functionality, an additional context item has been added, `implicitTimezone`.
```
rubensworks, line 363:
```suggestion
For the extended date functionality, an additional context item has been added: `implicitTimezone`.
```

rubensworks, line 363: Can we move this paragraph to the config section of the README? (and add `implicitTimezone` in the snippet there)

README.md
```diff
@@ -353,6 +353,17 @@ which you can provide as `baseIRI: string` to the config.
...
+The choice was made to default to UTC time so the `implicitTimezone` does not change over time
+(soo it is not daylight saving time dependent).
```
rubensworks, line 365:
```suggestion
(i.e., it is not dependent on daylight saving time).
```

lib/evaluators/AsyncEvaluator.ts
```diff
@@ -23,8 +23,15 @@ export class AsyncEvaluator {
   private readonly evaluator: IExpressionEvaluator<E.Expression, Promise<E.TermExpression>>;
 
   public static completeContext(context: IAsyncEvaluatorContext): ICompleteAsyncEvaluatorContext {
+    const now = context.now || new Date(Date.now());
+    let defaultTimeZone = context.defaultTimeZone;
+    if (!defaultTimeZone) {
+      // We make use of wrong modulo implementation: (-90) % 60 = -30
```
rubensworks, line 29: Why is it wrong?

lib/evaluators/SyncEvaluator.ts
```diff
@@ -23,8 +23,15 @@ export class SyncEvaluator {
   private readonly evaluator: IExpressionEvaluator<E.Expression, E.TermExpression>;
 
   public static completeContext(context: ISyncEvaluatorContext): ICompleteSyncEvaluatorContext {
+    const now = context.now || new Date(Date.now());
+    let defaultTimeZone = context.defaultTimeZone;
+    if (!defaultTimeZone) {
+      // We make use of wrong modulo implementation: (-90) % 60 = -30
+      defaultTimeZone = { zoneHours: Math.floor(now.getTimezoneOffset() / 60),
```
rubensworks, line 30: Duplicated code; we probably will want to abstract this part.

test/spec/_data.ts
```diff
@@ -43,15 +43,16 @@ export function data() {
...
+    d1: dateTimeTyped('2010-06-21T11:28:01Z'),
+    d2: dateTimeTyped('2010-12-21T15:38:02-08:00'),
+    d3: dateTimeTyped('2008-06-20T23:59:00Z'),
+    d4: dateTimeTyped('2011-02-01T01:02:03'),
 
-    dr1: date('2010-06-21'),
-    dr2: date('2010-12-21'),
-    dr3: date('2008-06-20'),
-    dr4: date('2011-02-01'),
+    // TODO: remove this comment: note: this is an illegal data format! date must contain time part
```
rubensworks, line 51: TODO

rubensworks (changes requested):

README.md
jitsedesmet: Ah, seems like this is outdated too, we altered this to default to the timezone in `now`.

lib/evaluators/AsyncEvaluator.ts
jitsedesmet: Yes, so it's not really wrong since most programming languages work like this but the mathematical modulo operator would return a positive integer. 

Would you want me to remove this comment?

test/spec/_data.ts
jitsedesmet: Yes! Tnx!

jitsedesmet:
This config has been deprecated, should I remove this? https://github.com/comunica/sparqlee/blob/master/README.md?plain=1#L66 (in a different PR?)

README.md
rubensworks: Yep

lib/evaluators/AsyncEvaluator.ts
rubensworks: Perhaps yes, to avoid confusion

lib/evaluators/AsyncEvaluator.ts
```diff
@@ -35,6 +41,7 @@ export class AsyncEvaluator {
       exists: context.exists,
       aggregate: context.aggregate,
       bnode: context.bnode,
+      defaultTimeZone,
```
rubensworks, line 44: Let's inline to simplify things.

```suggestion
      defaultTimeZone: context.defaultTimeZone || extractTimeZone(now),
```

(also in the other file)

rubensworks (changes requested):
Great! Just one minor nit.

Before we merge, have you tried running the test suite (with this PR linked in comunica)? https://github.com/kasei/sparql-12/tree/xsd_datetime_duration/tests/xsd_functions
We should be able to run it in a similar way as the normal spec tests. (see spec script in package.json of query-sparql)

jitsedesmet:
I have not tried to run the spec tests, no.
When I run `node ../../node_modules/rdf-test-suite/bin/Runner.js spec/sparql-engine.js https://raw.githubusercontent.com/kasei/sparql-12/xsd_datetime_duration/tests/xsd_functions/manifest.ttl -c ../../.rdf-test-suite-cache/`

I get the following error:
```
Error: Could not parse the RDF serialization text/plain; charset=utf-8 on https://raw.githubusercontent.com/kasei/sparql-12/xsd_datetime_duration/tests/xsd_functions/manifest.ttl
    at Util.parseRdfRaw (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/rdf-test-suite/lib/Util.js:84:15)
    at Util.<anonymous> (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/rdf-test-suite/lib/Util.js:59:46)
    at Generator.next (<anonymous>)
    at fulfilled (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/rdf-test-suite/lib/Util.js:5:58)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

rubensworks:
That's weird, GH pages should provide the correct content type based on file extensions.
In any case, we'll probably have to add an extensions-based fallback to rdf-test-suite at some point (e.g. by making use of `rdf-parse`).

As a temp workaround, could you modify this file (https://github.com/rubensworks/rdf-test-suite.js/blob/master/lib/Util.ts#L82, the local JS version) and a hardcoded workaround to use the N3 parser if the URL ends with `.ttl`?
That should allow you to run the tests locally.

jitsedesmet:
I added the following on the location you mentioned:
```ts
if (baseIRI.endsWith('.ttl')) {
  return data.pipe(new GeneralizedN3StreamParser({ baseIRI, format: 'text/turtle',  }));
}
if (baseIRI.endsWith('.srx')) {
  return data.pipe(new RdfXmlParser({ baseIRI }));
}
```

This makes me able to run the tests, but all tests fail. When manually checking the bindings Comunica things seem correct (I did this for one test), the tests suite however tells me the result is wrong. I suppose the `RdfXmlParser` is not the correct one? 

For completeness I add the log:  (the result file is: https://github.com/kasei/sparql-12/blob/xsd_datetime_duration/tests/xsd_functions/compare_duration-01.srx)
```
✖ compare xsd:duration values 01
  This tests the equality operator on xsd:duration values
  Error: Invalid query evaluation

  Query:

PREFIX xsd: <http://www.w3.org/2001/XMLSchema#>
SELECT ?id ?eq WHERE {
	VALUES (?id ?l ?r) {
		(1 "P1Y"^^xsd:duration "P1Y"^^xsd:duration)
		(2 "P1Y"^^xsd:duration "P12M"^^xsd:duration)
		(3 "P1Y"^^xsd:duration "P365D"^^xsd:duration)
		(4 "P0Y"^^xsd:duration "PT0S"^^xsd:duration)
		(5 "P1D"^^xsd:duration "PT24H"^^xsd:duration)
		(6 "P1D"^^xsd:duration "PT23H"^^xsd:duration)
		(7 "PT1H"^^xsd:duration "PT60M"^^xsd:duration)
		(8 "PT1H"^^xsd:duration "PT3600S"^^xsd:duration)
		(9 "-P1Y"^^xsd:duration "P1Y"^^xsd:duration)
		(10 "-P0Y"^^xsd:duration "PT0S"^^xsd:duration)
	}
	BIND(?l = ?r AS ?eq)
}


  Data links: 

  Result Source: https://raw.githubusercontent.com/kasei/sparql-12/xsd_datetime_duration/tests/xsd_functions/compare_duration-01.srx

  Expected: [QueryResultQuads: [
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#sparql",
    "graph": ""
  },
  {
    "subject": "_:df_10_1",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#variable",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#head",
    "object": "_:df_10_1",
    "graph": ""
  },
  {
    "subject": "_:df_10_2",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#variable",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#head",
    "object": "_:df_10_2",
    "graph": ""
  },
  {
    "subject": "_:df_10_3",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_3",
    "graph": ""
  },
  {
    "subject": "_:df_10_4",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_3",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_4",
    "graph": ""
  },
  {
    "subject": "_:df_10_5",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_3",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_5",
    "graph": ""
  },
  {
    "subject": "_:df_10_6",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_6",
    "graph": ""
  },
  {
    "subject": "_:df_10_7",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_6",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_7",
    "graph": ""
  },
  {
    "subject": "_:df_10_8",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_6",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_8",
    "graph": ""
  },
  {
    "subject": "_:df_10_9",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_9",
    "graph": ""
  },
  {
    "subject": "_:df_10_10",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_9",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_10",
    "graph": ""
  },
  {
    "subject": "_:df_10_11",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_9",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_11",
    "graph": ""
  },
  {
    "subject": "_:df_10_12",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_12",
    "graph": ""
  },
  {
    "subject": "_:df_10_13",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_12",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_13",
    "graph": ""
  },
  {
    "subject": "_:df_10_14",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_12",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_14",
    "graph": ""
  },
  {
    "subject": "_:df_10_15",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_15",
    "graph": ""
  },
  {
    "subject": "_:df_10_16",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_15",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_16",
    "graph": ""
  },
  {
    "subject": "_:df_10_17",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_15",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_17",
    "graph": ""
  },
  {
    "subject": "_:df_10_18",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_18",
    "graph": ""
  },
  {
    "subject": "_:df_10_19",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_18",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_19",
    "graph": ""
  },
  {
    "subject": "_:df_10_20",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_18",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_20",
    "graph": ""
  },
  {
    "subject": "_:df_10_21",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_21",
    "graph": ""
  },
  {
    "subject": "_:df_10_22",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_21",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_22",
    "graph": ""
  },
  {
    "subject": "_:df_10_23",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_21",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_23",
    "graph": ""
  },
  {
    "subject": "_:df_10_24",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_24",
    "graph": ""
  },
  {
    "subject": "_:df_10_25",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_24",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_25",
    "graph": ""
  },
  {
    "subject": "_:df_10_26",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_24",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_26",
    "graph": ""
  },
  {
    "subject": "_:df_10_27",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_27",
    "graph": ""
  },
  {
    "subject": "_:df_10_28",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_27",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_28",
    "graph": ""
  },
  {
    "subject": "_:df_10_29",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_27",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_29",
    "graph": ""
  },
  {
    "subject": "_:df_10_30",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#result",
    "graph": ""
  },
  {
    "subject": "_:df_10_0",
    "predicate": "http://www.w3.org/2005/sparql-results#results",
    "object": "_:df_10_30",
    "graph": ""
  },
  {
    "subject": "_:df_10_31",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_30",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_31",
    "graph": ""
  },
  {
    "subject": "_:df_10_32",
    "predicate": "http://www.w3.org/1999/02/22-rdf-syntax-ns#type",
    "object": "http://www.w3.org/2005/sparql-results#literal",
    "graph": ""
  },
  {
    "subject": "_:df_10_30",
    "predicate": "http://www.w3.org/2005/sparql-results#binding",
    "object": "_:df_10_32",
    "graph": ""
  }
]]

  Got: 
 [QueryResultBindings:
    Variables: [
  "?id",
  "?eq"
]
    Bindings:  [
  {
    "?id": {
      "termType": "Literal",
      "value": "1",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "true",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "2",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "true",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "3",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "false",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "4",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "true",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "5",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "true",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "6",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "false",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "7",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "true",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "8",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "true",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "9",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "false",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  },
  {
    "?id": {
      "termType": "Literal",
      "value": "10",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#integer"
      }
    },
    "?eq": {
      "termType": "Literal",
      "value": "true",
      "language": "",
      "datatype": {
        "termType": "NamedNode",
        "value": "http://www.w3.org/2001/XMLSchema#boolean"
      }
    }
  }
]
]
```

rubensworks:
`srx` is SPARQL/XML, not RDF/XML.

jitsedesmet:
Alright, yes; I added a case on: https://github.com/rubensworks/rdf-test-suite.js/blob/master/lib/testcase/sparql/TestCaseQueryEvaluation.ts#L44

```ts
if (url.endsWith('.srx')) {
  contentType = 'application/sparql-results+xml';
}
```

jitsedesmet:
Okey. Some things were still wrong (Oh how hard it is to copy spec tests correctly :'( )
These issues have been resolved. 3 spec tests still fail since they are related to the new `ADJUST` that has not been implemented. 
So, I think this is ready to be merged  :D  *excided*

```
ode ../../../rdf-test-suite.js/bin/Runner.js spec/sparql-engine.js https://raw.githubusercontent.com/kasei/sparql-12/xsd_datetime_duration/tests/xsd_functions/manifest.ttl -c ../../.rdf-test-suite-cache/ 
✔ compare xsd:duration values 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#compare_duration01) 46.032811ms
✔ compare xsd:yearMonthDuration values 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#compare_yearMonthDuration01) 11.421593ms
✔ compare xsd:dayTimeDuration values 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#compare_dayTimeDuration01) 5.038684ms
✔ compare xsd:date values 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#compare_time01) 6.430583ms
✔ extract xsd:date components 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#extract_date01) 3.986147ms
✔ extract xsd:time components 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#extract_time01) 3.739652ms
✖ xsd:dateTime timezone adjustment 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#adjust_dateTime01)
✖ xsd:date timezone adjustment 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#adjust_date01)
✖ xsd:time timezone adjustment 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#adjust_time01)
✔ xsd:dateTime, xsd:date, xsd:time subtraction 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#dateTime_subtract01) 3.604079ms
✔ xsd:yearMonthDuration addition 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#duration_yearMonth_add01) 8.030167ms
✔ xsd:dayTimeDuration addition 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#duration_dayTime_add01) 3.434092ms
✔ xsd:yearMonthDuration subtraction 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#duration_yearMonth_subtract01) 2.673866ms
✔ xsd:dayTimeDuration subtraction 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#duration_dayTime_subtract01) 4.500485ms
✔ xsd:date construction 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#constructor_date01) 3.267333ms
✔ xsd:date construction 02 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#constructor_date02) 3.607592ms
✔ xsd:time construction 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#constructor_time01) 4.17699ms
✔ xsd:time construction 02 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#constructor_time02) 2.764017ms
✔ xsd:duration construction 01 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#constructor_duration01) 4.603016ms
✔ xsd:duration construction 02 (http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#constructor_duration02) 3.522936ms

✖ xsd:dateTime timezone adjustment 01
  This tests ability to change the timezone of an xsd:dateTime value
  Error: Parse error on line 10:
...:dateTime)	}	BIND(ADJUST(?d, ?tz) AS ?
---------------------^
Expecting 'IRIREF', 'PNAME_NS', 'VAR', '(', 'INTEGER', '!', '<<', '+', '-', 'FUNC_ARITY0', 'FUNC_ARITY1', 'FUNC_ARITY2', 'IF', 'BOUND', 'BNODE', 'EXISTS', 'COUNT', 'FUNC_AGGREGATE', 'GROUP_CONCAT', 'DECIMAL', 'DOUBLE', 'BOOLEAN', 'STRING_LITERAL1', 'STRING_LITERAL2', 'STRING_LITERAL_LONG1', 'STRING_LITERAL_LONG2', 'INTEGER_POSITIVE', 'DECIMAL_POSITIVE', 'DOUBLE_POSITIVE', 'INTEGER_NEGATIVE', 'DECIMAL_NEGATIVE', 'DOUBLE_NEGATIVE', 'PNAME_LN', 'NOT', 'CONCAT', 'COALESCE', 'SUBSTR', 'REGEX', 'REPLACE', got 'a'
    at Parser.parseError (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/lib/SparqlParser.js:674:21)
    at Parser.parse (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/lib/SparqlParser.js:741:22)
    at Parser.parser.parse (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/sparql.js:36:37)
    at ActorQueryParseSparql.run (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.js:23:37)
    at ActorQueryParseSparql.runObservable (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/core/lib/Actor.js:57:29)
    at MediatorRace.mediate (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/core/lib/Mediator.js:79:22)
    at async QueryEngine.queryOrExplain (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/lib/QueryEngineBase.js:104:38)
    at async QueryEngine.query (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/lib/QueryEngineBase.js:42:24)
    at async Object.queryLdf (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/spec/sparql-engine-base.js:14:22)
  More info: http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#adjust_dateTime01


✖ xsd:date timezone adjustment 01
  This tests ability to change the timezone of an xsd:date value
  Error: Parse error on line 9:
...^xsd:date)	}	BIND(ADJUST(?d, ?tz) AS ?
---------------------^
Expecting 'IRIREF', 'PNAME_NS', 'VAR', '(', 'INTEGER', '!', '<<', '+', '-', 'FUNC_ARITY0', 'FUNC_ARITY1', 'FUNC_ARITY2', 'IF', 'BOUND', 'BNODE', 'EXISTS', 'COUNT', 'FUNC_AGGREGATE', 'GROUP_CONCAT', 'DECIMAL', 'DOUBLE', 'BOOLEAN', 'STRING_LITERAL1', 'STRING_LITERAL2', 'STRING_LITERAL_LONG1', 'STRING_LITERAL_LONG2', 'INTEGER_POSITIVE', 'DECIMAL_POSITIVE', 'DOUBLE_POSITIVE', 'INTEGER_NEGATIVE', 'DECIMAL_NEGATIVE', 'DOUBLE_NEGATIVE', 'PNAME_LN', 'NOT', 'CONCAT', 'COALESCE', 'SUBSTR', 'REGEX', 'REPLACE', got 'a'
    at Parser.parseError (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/lib/SparqlParser.js:674:21)
    at Parser.parse (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/lib/SparqlParser.js:741:22)
    at Parser.parser.parse (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/sparql.js:36:37)
    at ActorQueryParseSparql.run (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.js:23:37)
    at ActorQueryParseSparql.runObservable (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/core/lib/Actor.js:57:29)
    at MediatorRace.mediate (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/core/lib/Mediator.js:79:22)
    at async QueryEngine.queryOrExplain (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/lib/QueryEngineBase.js:104:38)
    at async QueryEngine.query (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/lib/QueryEngineBase.js:42:24)
    at async Object.queryLdf (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/spec/sparql-engine-base.js:14:22)
  More info: http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#adjust_date01


✖ xsd:time timezone adjustment 01
  This tests ability to change the timezone of an xsd:time value
  Error: Parse error on line 10:
...^xsd:time)	}	BIND(ADJUST(?d, ?tz) AS ?
---------------------^
Expecting 'IRIREF', 'PNAME_NS', 'VAR', '(', 'INTEGER', '!', '<<', '+', '-', 'FUNC_ARITY0', 'FUNC_ARITY1', 'FUNC_ARITY2', 'IF', 'BOUND', 'BNODE', 'EXISTS', 'COUNT', 'FUNC_AGGREGATE', 'GROUP_CONCAT', 'DECIMAL', 'DOUBLE', 'BOOLEAN', 'STRING_LITERAL1', 'STRING_LITERAL2', 'STRING_LITERAL_LONG1', 'STRING_LITERAL_LONG2', 'INTEGER_POSITIVE', 'DECIMAL_POSITIVE', 'DOUBLE_POSITIVE', 'INTEGER_NEGATIVE', 'DECIMAL_NEGATIVE', 'DOUBLE_NEGATIVE', 'PNAME_LN', 'NOT', 'CONCAT', 'COALESCE', 'SUBSTR', 'REGEX', 'REPLACE', got 'a'
    at Parser.parseError (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/lib/SparqlParser.js:674:21)
    at Parser.parse (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/lib/SparqlParser.js:741:22)
    at Parser.parser.parse (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/node_modules/sparqljs/sparql.js:36:37)
    at ActorQueryParseSparql.run (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-query-parse-sparql/lib/ActorQueryParseSparql.js:23:37)
    at ActorQueryParseSparql.runObservable (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/core/lib/Actor.js:57:29)
    at MediatorRace.mediate (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/core/lib/Mediator.js:79:22)
    at async QueryEngine.queryOrExplain (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/lib/QueryEngineBase.js:104:38)
    at async QueryEngine.query (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/lib/QueryEngineBase.js:42:24)
    at async Object.queryLdf (/home/jitsedesmet/Documents/werken/22-jaar-comunica/comunica/packages/actor-init-query/spec/sparql-engine-base.js:14:22)
  More info: http://www.w3.org/2009/sparql/docs/tests/data-sparql11/functions/manifest#adjust_time01

✖ 17 / 20 tests succeeded!
```

rubensworks:
This is really amazing work @jitsedesmet! Preparing a new release now...

Could you look into making a PR to Comunica to include this change?
We'll also have to update rdf-test-suite with your changes, as we need to run these tests in the CI as well.
(we'll have to run with the `-e` option, to not let the CI fail because of those 3 failing tests)
Once all of that is done, can you open a PR to https://github.com/w3c/sparql-12/blob/main/SEP/SEP-0002/sep-0002.md to add Comunica to the list of supported implementation (perhaps with an `ADJUST` disclaimer).

PR #165, opened 2023-01-18
Fix cache search problem

jitsedesmet:
Running the spec tests of the Comunica repo showed something was wrong in the cache search for functions hanging multiple arity. This PR is a fix for this problem.

### DIFF
--- a/lib/functions/OverloadTree.ts
+++ b/lib/functions/OverloadTree.ts
@@ -78,7 +78,7 @@ export class OverloadTree {
   public search(args: E.TermExpression[], superTypeProvider: ISuperTypeProvider,
     functionArgumentsCache: FunctionArgumentsCache):
     ImplementationFunction | undefined {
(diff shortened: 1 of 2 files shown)


rubensworks (changes requested):
Looks good! Could you just add a unit test to validate this functionality?

lib/functions/OverloadTree.ts
```diff
@@ -131,7 +130,7 @@ export class OverloadTree {
     for (const term of args) {
       const literalExpression = isLiteralTermExpression(term);
       const key = literalExpression ? literalExpression.dataType : term.termType;
-      cache.cache = {};
```
jitsedesmet, line 134: I altered this line because it basically reseted some parts of the cache sometimes. (I re-checked whether the cache size still had manageable size and all seemed fine :)

rubensworks (approved):

Issue #167, opened 2023-02-23
Equality check Error should not be thrown


danielbeeke:
I think the error on https://github.com/comunica/sparqlee/blob/master/lib/functions/RegularFunctions.ts#L111 should not be thrown when the terms are equal.

 See: https://www.w3.org/TR/rdf-concepts/#section-Literal-Equality

> produces a type error if the arguments are both literal but are not the same RDF term

Could it be we just have to add `!val && ` to the if statement?

See https://github.com/comunica/comunica/issues/1146

Issue #2, opened 2018-09-27
Plain literal behaviour


wschella:
### Context

In the SPARQL spec, more spefically [operand data types](https://www.w3.org/TR/sparql11-query/#operandDataTypes) and [operator mapping] there exist the concepts of:
- plain literals: one of
  - simple literals 
  - language tagged literals
- typed literals with `xsd:string type`

### Problem

The distinction is important, as the string operators only apply to apply to typed literals, and not the plain ones. For example, following expressions are **NOT** equivalent:
`"foo"^^xsd:string != "bar"^^xsd:string` and `"foo" != "bar"`.
The first one will evaluate true, the second one should error, since the operator is not defined. A solution to this is to allow simple literals in problematic operators.

A bigger problem however is that the return type the some functions (mainly functions on strings) is dependent on the original type, e.g. a simple literal should be returned when the (first) argument was a simple literal. This is however impossible.

However: the [RDF.js](https://github.com/rdfjs/representation-task-force/blob/master/interface-spec.md) spec hides the difference between **simple** and **typed** literals, as it automatically adds the `xsd:string` data type. 

### Proposal

Concluded from the discussion on the [RDF.js repo](https://github.com/rdfjs/representation-task-force/issues/127), we want to consider simple literals as being typed with `xsd:string`. By extension, the same goes for language tagged literals and `rdf:langString`.

Any untyped literals we get as input we could type ourselves.

This will give issues with spec tests.

wschella:
Behavior mentioned above is implemented, but this should still be documented somewhere.

wschella:
This does not take into consideration:

> 17.4.3.1.3 String Literal Return Type

> Functions that return a string literal do so with the string literal of the same kind as the first argument (simple literal, plain literal with same language tag, xsd:string). This includes SUBSTR, STRBEFORE and STRAFTER.

> The function CONCAT returns a string literal based on the details of all its arguments.

wschella:
This behavior is implemented and final.

PR #168, opened 2023-02-23
Adding the val to the statement of the literal equality check

danielbeeke:
Here is the issue:
https://github.com/comunica/sparqlee/issues/167

### DIFF
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -94,6 +94,10 @@ const equality = {
   overloads: declare(C.RegularOperator.EQUAL)
     .numberTest(() => (left, right) => left === right)
     .stringTest(() => (left, right) => left.localeCompare(right) === 0)
+    .set(
+      [ 'literal', 'literal' ],
+      () => ([ left, right ]) => bool(RDFTermEqual(left, right)),
+    )
     .booleanTest(() => (left, right) => left === right)
     .dateTimeTest(() => (left, right) => left.getTime() === right.getTime())
     .set(
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -107,7 +107,7 @@ function RDFTermEqual(_left: Term, _right: Term): boolean {
   const left = _left.toRDF();
   const right = _right.toRDF();
   const val = left.equals(right);
-  if ((left.termType === 'Literal') && (right.termType === 'Literal')) {
+  if (!val && (left.termType === 'Literal') && (right.termType === 'Literal')) {
     throw new Err.RDFEqualTypeError([ _left, _right ]);
   }
   return val;
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -95,8 +95,11 @@ const equality = {
     .numberTest(() => (left, right) => left === right)
     .stringTest(() => (left, right) => left.localeCompare(right) === 0)
     .set(
-      [ 'literal', 'literal' ],
-      () => ([ left, right ]) => bool(RDFTermEqual(left, right)),
+      [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
+      () => ([ left, right ]) => {
+        console.log(left)
+        return bool(RDFTermEqual(left, right))
+      },
     )
     .booleanTest(() => (left, right) => left === right)
     .dateTimeTest(() => (left, right) => left.getTime() === right.getTime())
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -96,7 +96,8 @@ const equality = {
     .stringTest(() => (left, right) => left.localeCompare(right) === 0)
     .set(
       [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
-      () => ([ left, right ]) => bool(RDFTermEqual(left, right)),
+      () => ([ left, right ]: E.LangStringLiteral[]) => bool(left.str() === right.str() &&
+        left.language === right.language),
     )
     .booleanTest(() => (left, right) => left === right)
     .dateTimeTest(() => (left, right) => left.getTime() === right.getTime())
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -96,10 +96,7 @@ const equality = {
     .stringTest(() => (left, right) => left.localeCompare(right) === 0)
     .set(
       [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
-      () => ([ left, right ]) => {
-        console.log(left)
-        return bool(RDFTermEqual(left, right))
-      },
+      () => ([ left, right ]) => bool(RDFTermEqual(left, right)),
     )
     .booleanTest(() => (left, right) => left === right)
     .dateTimeTest(() => (left, right) => left.getTime() === right.getTime())
--- a/test/integration/functions/op.equality.test.ts
+++ b/test/integration/functions/op.equality.test.ts
@@ -65,9 +65,7 @@ describe('evaluation of \'=\'', () => {
         aaa   aaa   = true
         aaa   bbb   = false
         "test1" "test2" = false
-      `,
-      errorTable: `
-        "test1"@en "test2"@en = 'Equality test for literals with unsupported datatypes'
+        "test1"@en "test2"@en = false
       `,
     });
   });
--- a/test/integration/functions/op.equality.test.ts
+++ b/test/integration/functions/op.equality.test.ts
@@ -106,6 +106,7 @@ describe('evaluation of \'=\'', () => {
         1 <http://example.com> = false
       `,
       errorTable: `
+        "test1"@en = "test2"@en
         1 true = 'Equality test for literals with unsupported datatypes'
         1 aaa = 'Equality test for literals with unsupported datatypes'
         1 earlyN = 'Equality test for literals with unsupported datatypes'
--- a/test/integration/functions/op.equality.test.ts
+++ b/test/integration/functions/op.equality.test.ts
@@ -58,11 +58,15 @@ describe('evaluation of \'=\'', () => {
     runTestTable({
       ...config,
       testTable: `
+        "test1"@en "test1"@en = true
         empty empty = true
         empty aaa   = false
         aaa   aaa   = true
         aaa   bbb   = false
       `,
+      errorTable: `
+        "test1"@en "test2"@en = 'Equality test for literals with unsupported datatypes'
+      `
     });
   });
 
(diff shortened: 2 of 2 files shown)


jitsedesmet:
This looks good to me! Would it be possible to add a test, so we don't run into this issue again?

danielbeeke:
Yes, Ill create a test

danielbeeke:
Ah I made a mistake with the commit. I used a wrong username. I have now signed it with both GitHub accounts.

jitsedesmet (commented):
Alright, can you just add the same tests without the language tag too? (Not that it matters, just because the original issue mentioned this too). As soon as this is done, this should be good to go!

test/integration/functions/op.equality.test.ts
```diff
@@ -58,10 +58,16 @@ describe('evaluation of \'=\'', () => {
     runTestTable({
       ...config,
       testTable: `
+        "test1"@en "test1"@en = true
+        "test1" "test1" = true
         empty empty = true
         empty aaa   = false
         aaa   aaa   = true
         aaa   bbb   = false
+        "test1" "test2" = false
```
jitsedesmet, line 67: Is it normal that the case without language strings return `false` and with language string fails? Both are literals right? I'm a little confused here. 

Additionally, the original [Communica issue](https://github.com/comunica/comunica/issues/1146) states:
> Any FILTER on ?o in the example fails (with or without language string, = or !=).

So there it seems to also fail without language strings?  (This issue might refer to `"test" = "test"@en`, if this is the case, a test regarding this case is in place.)

danielbeeke: ## The current situation with Sparqlee:

Treats as Literal in the case of equality checks:
- "test"
- "test"@en

Treats as strings and checks them here: https://github.com/comunica/sparqlee/blob/master/lib/functions/RegularFunctions.ts#L96

- test

## What should we do? 

I think the consensus in other issues is that Literals and strings inside Sparql should be treated as Literals. 

See:

- https://github.com/comunica/sparqlee/issues/2
- https://github.com/rdfjs/data-model-spec/issues/127#issuecomment-385320677
- 
Especially the last part of the last link seems important:

> In 2014, the RDF 1.1 WG decided that the pain of maintaining a distinction between simple literal and xsd:string-typed literal exceeded the pain of migration to a model where everything had a datatype. In the latter model, parsers that appeared to have a simple literal form were expected to produce literals with type xsd:string. You are experiencing some of that migration pain. Apologies for not getting this right in 1999.

That would mean we have to:

- Treat `test` and `"test"@en` the same AND both should throw errors in case of a comparison with an other string that is not equal.

This mean changing internals of Sparqlee to pick up 'plain strings' as Literals.

@jitsedesmet What are your thoughts?

jitsedesmet: Yes, I think this means we should just handle simple literals as xsd:string-typed literals as soon as they enter sparqlee. I'll ask for @rubenworks his opinion. If he gives his thumbs up, I suggest you look at the files in the transformers directory.

rubensworks: > Is it normal that the case without language strings return false and with language string fails? Both are literals right? I'm a little confused here.

Note that the test here checks equality for `"test1"` and `"test2"`, which are different strings, hence the failure.

> I think this means we should just handle simple literals as xsd:string-typed literals as soon as they enter sparqlee

Indeed. AFAIK, this is how things work today.
Even the RDF/JS data model has been designed like this. An untyped literal string will default to the string datatype.

> Treat test and "test"@en the same AND both should throw errors in case of a comparison with an other string that is not equal.

No, `"test"` and `"test"@en` are not the same. The one has a string datatype without language, and the other is a language-tagged-string with the `en` language.

danielbeeke: > No, "test" and "test"@en are not the same. The one has a string datatype without language, and the other is a language-tagged-string with the en language.

Sure, I agree, I meant something different with my comment. I meant:

Treat all the following cases 'the same', the same equality check:

_Should pass:_
test1 = test1
"test1"@en = "test1"@en

_Should not pass:_
test1 = test2
"test1"@en = "test2"@en
"test1"@en = "test1"
"test1"@en = test1

The point being that they all undergo the same equality check, 
which at the moment in Sparqlee is not the case. 

I tested the code and plain string equality seemed to happen here: https://github.com/comunica/sparqlee/blob/master/lib/functions/RegularFunctions.ts#L96

Term equality went through this: https://github.com/comunica/sparqlee/blob/master/lib/functions/RegularFunctions.ts#L106

rubensworks: > Treat all the following cases 'the same', the same equality check:

Yep, that's correct, according to the [spec](https://www.w3.org/TR/rdf-concepts/#section-Literal-Equality).

> The point being that they all undergo the same equality check,
which at the moment in Sparqlee is not the case.

I suspect the datatype and language checks would be part of [this line](https://github.com/comunica/sparqlee/blob/master/lib/functions/RegularFunctions.ts#L109).

danielbeeke: > Yes, I think this means we should just handle simple literals as xsd:string-typed literals as soon as they enter sparqlee. I'll ask for @rubenworks his opinion. If he gives his thumbs up, I suggest you look at the files in the transformers directory.

Shall I proceed?

jitsedesmet: Yes.
However, the approach I suggested regarding `TermTransfromer` might not be correct. 
I'll try to summarize and reason how we can proceed.

The spec tells us `"test"` should become `"test"^^xsd:string` as soon as it is parsed.
Similarly, `"test"@en` should become `"test"@en^^rdf:langString` or `"test"@en^^xsd:string`.
(The docs should be read for this).
I think it'll be `rdf:langString` because of https://www.w3.org/TR/rdf11-concepts/#h3_section-Graph-Literal, but it should be checked.
I think @rubensworks had found an official document regarding the transformation of plain literals.
I don't know if this is handled by sparqlee or not, and I don't know whether we use a package for this.

That being said, we should look what behavior the spec of equality describes.
Equality is described in: https://www.w3.org/TR/sparql11-query/#func-isLiteral#OperatorMapping 
For strings this is: https://www.w3.org/TR/xpath-functions/#func-compare
For language string I have not found a custom equality function, meaning it should be checked with the term equal: https://www.w3.org/TR/sparql11-query/#func-RDFterm-equal

If all of this is right, I think the implementation in the current PR is right. (?) I will however first check if I find anything for the language equality.

jitsedesmet: Alright. Since I don't find any special treatment for languageStrings, this implementation should be correct and ready to merge.

jitsedesmet (approved):
This is ready to merge.

Summarization:
It might be weird that an error is thrown when comparing 2 non-equal languageStrings and not when comparing 2 non-equal strings, but this should be the behavior according to the spec. I tried to list the reason in a comment here: https://github.com/comunica/sparqlee/pull/168#discussion_r1120072061

rubensworks:
> Summarization:
It might be weird that an error is thrown when comparing 2 non-equal languageStrings and not when comparing 2 non-equal strings, but this should be the behavior according to the spec. I tried to list the reason in a comment here: https://github.com/comunica/sparqlee/pull/168#discussion_r1120072061

It's a bit weird indeed, however correct according to the spec.

Since we all agree it's weird, I suggest that we go beyond the spec, and properly support the comparison on rdf:langString's as well, which should require minimal effort AFAICS.
Since SPARQL allows [operator extensibility](https://www.w3.org/TR/sparql11-query/#operatorExtensibility) for undefined operations, this will still be spec-compliant.

jitsedesmet:
I agree. And this should go fluently, by using the `set` command after setting the stringy overload on the equality.
Between these lines:  https://github.com/comunica/sparqlee/blob/master/lib/functions/RegularFunctions.ts#L96-L97

lib/functions/RegularFunctions.ts
```diff
@@ -94,6 +94,10 @@ const equality = {
   overloads: declare(C.RegularOperator.EQUAL)
     .numberTest(() => (left, right) => left === right)
     .stringTest(() => (left, right) => left.localeCompare(right) === 0)
+    .set(
+      [ 'literal', 'literal' ],
+      () => ([ left, right ]) => bool(RDFTermEqual(left, right)),
+    )
```
jitsedesmet, line 100: It would be more something like this. Note that this is more pseudocode than anything else :)

```suggestion
    .set(
      [ TypeUrl.LanguageString, TypeUrl.LanguageString ],
      () => ([ left, right ]) => bool(left.str() === right.str() && left.language() === right.language()),
    )
```

danielbeeke: Thanks, I will have look, I understand more now

danielbeeke: What do you think of the current code? I still forward the checking to RDFTermEqual. It would then pass to something like https://github.com/rdfjs/N3.js/blob/main/src/N3DataFactory.js#L111

danielbeeke: Ahh I see it goes through this one:

https://github.com/rubensworks/rdf-data-factory.js/blob/master/lib/Literal.ts#L33

Same idea.

lib/functions/RegularFunctions.ts
```diff
@@ -94,6 +94,10 @@ const equality = {
   overloads: declare(C.RegularOperator.EQUAL)
     .numberTest(() => (left, right) => left === right)
     .stringTest(() => (left, right) => left.localeCompare(right) === 0)
+    .set(
+      [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
+      () => ([ left, right ]) => bool(RDFTermEqual(left, right)),
```
jitsedesmet, line 99: When using the rdf term equal it would also check if the datatype is the same. Our type system is able to handle subtypes. I think it would be better to support that feature. (This just means you need to check value and language by yourself instead of using RDFTermEqual)

jitsedesmet: Oh, I see you where commenting on my previous, sorry :sweat_smile:

danielbeeke: okay, will do

danielbeeke: Is the error needed when two TypeURL.RDF_LANG_STRINGs are not equal?

If that is the case it wonder if RDFTermEqual might be fine.

jitsedesmet: I don't know if I understand your question. So we will implement a languageString equality so that it does not throw an error when they are not equal. Instead, we want it to return false. 

We can not use RDFTermEqual since it will check if the datatype is also equal. However, Sparqlee can discover subTypes. In other words, if I define `myLangStr`, a subtype of `languageString`, it should still work i the equality. (As in, this is how all other equality functions also work)

danielbeeke: > So we will implement a languageString equality so that it does not throw an error when they are not equal. Instead, we want it to return false.

Ah okay, yes that would be an answer to the question. So I also have change the test that changed before.
I am on it.

jitsedesmet (approved):
Looks good to me! @rubensworks Can you merge this?

rubensworks (approved):
Nice work @danielbeeke!

rubensworks:
Released as 2.2.1.

danielbeeke:
Thank you @jitsedesmet and @rubensworks for persisting with me and reviewing!

PR #172, opened 2023-03-24
Fix/distinct on aggregate

### LINKED ISSUE #151: DISTINCT is not supported yet on aggregates
Aggregate operators do not seems to support the distinct option yet.

Here is a PR that highlight the problem (see the added TODO) #152

jitsedesmet:
This PR aims to fix #151.

I fear this issue will become rather large when we want to solve this issue in an appropriate manner.
(This because the typing of the aggregate function lacks information to easilly implement this)


jitsedesmet:
These first few commits result in a working implementation for the count aggregate. However, it also results in a major speed reduction since each binding in a list of bindings is evaluated seperatly instead of all at once. Two options here that I see:
1) The `AggregateEvaluators` know when they are handling a wildcard Expression and only than use this slow splitting ) behaviour
2) We change the returntype of our evaluators to return a list of terms.

jitsedesmet:
Seems like I was looking in a wrong direction. I was also trying to implement the wildcard `*`, but the specs state that wildcars is only possible on the count aggregator https://www.w3.org/TR/sparql11-query/#rAggregate

(that's probably why I only saw how I could implement it for that aggregator hehe)

jitsedesmet:
This PR is now ready. Note, it handles the requested task of fixing distinct. It however does not yet fix the problem when using wildcards. I will try to fix this problem next, in a sepperate PR.

jitsedesmet:
A hashing approach might me beter here but will result in more memory usage. @rubensworks I leave it up to you too decide on that. :)

jitsedesmet:
I managed to also fix the wildcards in a somewhat easy way since we only need to handle them for count I think this approach is acceptable.

lib/aggregators/BaseAggregator.ts
```diff
@@ -43,9 +49,59 @@ export abstract class BaseAggregator<State> {
...
+
+  protected canSkip(state: IBaseState, term: RDF.Term | undefined, variable: string): boolean {
+    return this.distinct && (state.seen[variable] || []).some((t: RDF.Term) => t.equals(term));
+  }
+
+  protected addSeen(seen: Seen, term: RDF.Term | undefined,
+    variable: string): Seen {
+    return this.distinct ?
+      {
+        ...seen,
```
rubensworks, line 85: This operation may get a bit too expensive due to object re-recreation and copying.

I suggest to make use of a mutable `Map<string, Set<string>>`, which maps variables to a set of stringified (using `rdf-string`) terms.
The stringified terms can then just be checked using `has` instead of iteration.

lib/aggregators/BaseAggregator.ts
```diff
@@ -7,7 +7,13 @@ import type { ITermTransformer } from '../transformers/TermTransformer';
 import { TypeAlias } from '../util/Consts';
 import { isSubTypeOf } from '../util/TypeHandling';
 
-export abstract class BaseAggregator<State> {
+export interface IBaseState {
+  seen: Seen;
+}
+
+type Seen = Record<string, (RDF.Term | undefined)[]>;
```
rubensworks, line 14:
```suggestion
type VariableValues = Record<string, (RDF.Term | undefined)[]>;
```

lib/aggregators/BaseAggregator.ts
```diff
@@ -7,7 +7,13 @@ import type { ITermTransformer } from '../transformers/TermTransformer';
 import { TypeAlias } from '../util/Consts';
 import { isSubTypeOf } from '../util/TypeHandling';
 
-export abstract class BaseAggregator<State> {
+export interface IBaseState {
+  seen: Seen;
```
rubensworks, line 11:
```suggestion
  distinctValuesPassed: Seen;
```

lib/evaluators/evaluatorHelpers/BaseAggregateEvaluator.ts
```diff
@@ -1,23 +1,31 @@
...
 
   protected constructor(expr: Algebra.AggregateExpression,
     sharedContext: ICompleteSharedContext, throwError?: boolean) {
     this.expression = expr;
     this.aggregator = new aggregators[<SetFunction> expr.aggregator](expr, sharedContext);
     this.throwError = throwError || false;
+    this.isWildcard = expr.expression.expressionType === Algebra.expressionTypes.WILDCARD;
+    if (this.isWildcard) {
+      this.wildcardAggregator = new WildcardCount(expr, sharedContext);
+      this.put = this.__initWildcard.bind(this);
```
rubensworks, line 27: Using `bind` should be avoided if possible, as it can cause obscure bugs.
Can we move these new methods into the `WildcardCount` class instead?

lib/aggregators/BaseAggregator.ts
```diff
@@ -43,9 +49,59 @@ export abstract class BaseAggregator<State> {
...
+    return this.distinct ?
+      {
+        ...seen,
+        [variable]: [ ...seen[variable] || [], term ],
+      } :
+      {};
+  }
+}
+
+export abstract class SimpleAggregator<SubState> extends BaseAggregator<SubState> {
```
rubensworks, line 92: I'm not a big fan of the `sub...` method names. Feels a bit hacky.

Instead of extension, I would suggest using object composition instead, which should solve this problem.

Concretely, the `BaseAggregator` can accept an `IAggregatorComponent` in its constructor, which has the old `BaseAggregator` interface.
Then, the `Count`, `Max`, ... classes can just be implementations of `IAggregatorComponent`.
This makes it possible to still use the old `init`, `put`, and `result` method names.

Instantiations of the different aggregators will then look something like this: `new BaseAggregator(new Count())`.
Since `BaseAggregator` won't be an abstract class anymore, we could even name it `Aggregator` now.

rubensworks (changes requested):

lib/evaluators/evaluatorHelpers/BaseAggregateEvaluator.ts
jitsedesmet: I agree. Right now, the aggregate evaluators use `binds`, I tried to work in the same fassion as what is already there. Getting rid of the `bind`s completly is not an option I think. I could reduce the amound of  times we use it.

I don't think we should fix all the binds in this PR. 
If I don't use a bind, the code stands out against the existing approuch. I don't know if this is bad?

lib/aggregators/BaseAggregator.ts
jitsedesmet: Perfect! I we're working with mutable types. I think we should stop passing the state as arguments to for example the `put` function.

lib/evaluators/evaluatorHelpers/BaseAggregateEvaluator.ts
jitsedesmet: I'll reflect back on this later. Maybe when altering the code in acordance to the next command, it will be easy to also change the bind behaviour.

jitsedesmet:
I started refactoring like you asked and got a little carried away. Imo this makes the aggregators way more readable. And less complex. The put function now has more cases to cover but since we established using the binds is confusing, this might not be a bad thing?

lib/aggregators/Aggregator.ts
```diff
@@ -0,0 +1,119 @@
...
+      }
+      this.variableValues.get(variable)!.add(term ? RdfString.termToString(term) : undefined);
+    }
+  }
+}
+
+/**
+ * A base aggregator that can handle distinct ignores wildcards.
+ */
+export class Aggregator extends baseAggregator {
```
rubensworks, line 94: Why do we need a separate `baseAggregator`?
Can we just make this the `Aggregator`?

lib/aggregators/Aggregator.ts
```diff
@@ -0,0 +1,119 @@
...
+    return undefined;
+  }
+
+  public constructor(expr: Algebra.AggregateExpression, protected sharedContext: ICompleteSharedContext) {
+    this.separator = expr.separator || ' ';
+    this.termTransformer = new TermTransformer(sharedContext.superTypeProvider);
+  }
+
+  protected termToNumericOrError(term: RDF.Term): E.NumericLiteral {
+    // TODO: Check behaviour
```
rubensworks, line 29: TODO

(consider adding a check in your editor that warn if there are TODO's before committing)

rubensworks (changes requested):
Great, I like the new components approach!

lib/aggregators/Aggregator.ts
jitsedesmet: Base Aggregator can make a distinction in variables, Aggregator can not, we could merge them into one and make variable a parameter with a default value of `''`. That might be cleaner?

lib/aggregators/Aggregator.ts
jitsedesmet: This TODO was already there. I just left it here. I can remove it since it is tested? (100% coverage?)

rubensworks: Ah ok. Let's remove then.

jitsedesmet:
Seems to work :D

jitsedesmet:
Seems like there still was a small issue. This should now be resolved.

jitsedesmet:
Retested it with Comunica, and the 3 tests suites pass.

rubensworks:
Released as 2.3.1

PR #175, opened 2023-04-17
feat: stringly typed equality

jitsedesmet:
This PR aims to fix: https://github.com/comunica/comunica/issues/1185.
I didn't add a separate test for this yet because I don't think this is core functionality.
If a test would be desired, I'd happily write one.

### DIFF
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -165,11 +165,10 @@ const equality = {
   arity: 2,
   overloads: declare(C.RegularOperator.EQUAL)
     .numberTest(() => (left, right) => left === right)
-    .stringTest(() => (left, right) => left.localeCompare(right) === 0)
     .set(
-      [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
-      () => ([ left, right ]: E.LangStringLiteral[]) => bool(left.str() === right.str() &&
-        left.language === right.language),
+      [ TypeAlias.SPARQL_STRINGLY, TypeAlias.SPARQL_STRINGLY ],
+      () => ([ left, right ]: (E.LangStringLiteral | E.StringLiteral)[]) => bool(left.str() === right.str() &&
+        left.language === right.language && left.dataType === right.dataType),
     )
     .booleanTest(() => (left, right) => left === right)
     .dateTimeTest(({ defaultTimeZone }) => (left, right) =>


lib/functions/RegularFunctions.ts
```diff
@@ -165,11 +165,10 @@ const equality = {
...
   overloads: declare(C.RegularOperator.EQUAL)
     .numberTest(() => (left, right) => left === right)
-    .stringTest(() => (left, right) => left.localeCompare(right) === 0)
     .set(
-      [ TypeURL.RDF_LANG_STRING, TypeURL.RDF_LANG_STRING ],
-      () => ([ left, right ]: E.LangStringLiteral[]) => bool(left.str() === right.str() &&
-        left.language === right.language),
+      [ TypeAlias.SPARQL_STRINGLY, TypeAlias.SPARQL_STRINGLY ],
+      () => ([ left, right ]: (E.LangStringLiteral | E.StringLiteral)[]) => bool(left.str() === right.str() &&
+        left.language === right.language && left.dataType === right.dataType),
```
jitsedesmet, line 171: Note: I added the datatype check here. But I'd need to check what the spec says about this.

rubensworks: Makes sense to me!

rubensworks:
> I didn't add a separate test for this yet because I don't think this is core functionality.
If a test would be desired, I'd happily write one.

Of course, we need to test everything! 😄

test/integration/functions/op.equality.test.ts
```diff
@@ -66,6 +66,7 @@ describe('evaluation of \'=\'', () => {
         aaa   bbb   = false
         "test1" "test2" = false
         "test1"@en "test2"@en = false
+        "test1" "test2"@en = false
```
rubensworks, line 69: Could you also add `"test1" "test1"@en = false` ?

lib/functions/RegularFunctions.ts
jitsedesmet: Yes, the only problem is, do we want this for our type discovery? What happens if we evaluate `"test"^^ex:myStr == "test"^^xsd:string` and we discovered that `ex:myStr` is a subtype of `xsd:string`? I think, in this case, the terms are equal? 

I think the spec says the same, but I am not 100% sure if I interoperate it correctly, could you verify?
In [Operand Data Types](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#operandDataTypes) the last sentence:
> SPARQL language extensions may treat additional types as being derived from XML schema datatypes.

I think this means: If you get `ex:myStr`, a subtype of `xsd:string`, you can use the following functions as if it was an `xsd:string`.

rubensworks: No, AFAIK, literals with unequal datatype are never equal.

The part of the spec you refer to talks about acceptable function args, for which subtypes may be considered.

jitsedesmet: I don't think this is true, although it makes sense... When checking for equality using [term equal](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#OperatorMapping), it is the case that both types need to be the same.
However, the [operator mapping part](https://www.w3.org/TR/2013/REC-sparql11-query-20130321/#OperatorMapping) just says subtype substitution needs to be handled. If subtype substitution needs to be handled, I think my example above should result in true.

rubensworks: Hmm, good catch, I think you're right.

So this statement is correct then:

> What happens if we evaluate "test"^^ex:myStr == "test"^^xsd:string and we discovered that ex:myStr is a subtype of xsd:string? I think, in this case, the terms are equal?

Do you think this will break with this PR?

jitsedesmet: As it is now, It will break, but I will make a commit later so everything is correct.

rubensworks (approved):
Very elegant solution!

rubensworks:
Released as 2.3.2

PR #176, opened 2023-04-22
refactor: use wrapper function for invalidLitteral protection

### LINKED ISSUE #115: Handle invalid lexical form better
@wschella I was wondering whether we can handle the invalidLexical form error from with the overloadTree?
Right now, the Builder enables this behavior from within its functions. This creates some tedious code, since every function added should provide all implementations of InvalidLexicalForm? This behavior makes some functions not function correctly, I think. For example, `"apple"^^xsd:integer + "0"^^xsd:integer` doesn't throw the right error. It throws implementation not found.

jitsedesmet:
This PR aims to fix #115 .

### DIFF
--- a/lib/expressions/Term.ts
+++ b/lib/expressions/Term.ts
@@ -328,18 +328,18 @@ export class YearMonthDurationLiteral extends Literal<Partial<IYearMonthDuration
 export class NonLexicalLiteral extends Literal<{ toString: () => 'undefined' }> {
   public constructor(
     typedValue: undefined,
-    public typeURL: string,
+    dataType: string,
     private readonly openWorldType: ISuperTypeProvider,
     strValue?: string,
     language?: string,
   ) {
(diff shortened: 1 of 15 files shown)


rubensworks:
I just have two questions on this:
- Does this have any impact on performance? (if so, how much)
- Do all spec tests (in comunica) still pass with this change?

jitsedesmet:
I checked the Comunica test suite last week and everything still seemed to be working.

As for performance, I posted an explanation in the related issue https://github.com/comunica/sparqlee/issues/115#issuecomment-1518646087 . It boils down to having a single if statement before each function. I think compared to the work done in the overloadTree, this should have minimal impact.

jitsedesmet:
I have tested again (to be sure), and can confirm the comunica tests still work (sorry, running these kill my laptop 60% of the time).

PR #161, opened 2022-11-27
feat/rdfstar

jeswr:
- feat: add support for sparql-star expressions
- WIP: testing sparqlStar
- sparqlstar operator tests passing
- chore: run lint:fix
- WIP: all sparqlstar operations

PR #179, opened 2023-06-23
Add SPARQL-star operators and comparison

rubensworks:
Partially built on top of @jeswr's WIP PR from #161.

@jitsedesmet I changed the ordering logic quite a bit (the old version had a fundamental flaw where `>` was considered the opposite to `<`), so it might be good for you to check if I didn't break anything major (all unit and spec tests still pass).

### DIFF
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -1,3 +1,4 @@
+import type * as RDF from '@rdfjs/types';
 import { BigNumber } from 'bignumber.js';
 import { sha1, sha256, sha384, sha512 } from 'hash.js';
 import { DataFactory } from 'rdf-data-factory';
@@ -7,9 +8,10 @@ import * as uuid from 'uuid';
 
 import type { ICompleteSharedContext } from '../evaluators/evaluatorHelpers/BaseExpressionEvaluator';
 import * as E from '../expressions';
+import type { Quad } from '../expressions';
 import { TermTransformer } from '../transformers/TermTransformer';
 import * as C from '../util/Consts';
-import { TypeAlias, TypeURL } from '../util/Consts';
+import { RegularOperator, TypeAlias, TypeURL } from '../util/Consts';
 import type { IDayTimeDurationRepresentation } from '../util/DateTimeHelpers';
 import {
   dayTimeDurationsToSeconds,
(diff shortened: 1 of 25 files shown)


rubensworks:
PR fails as this requires https://github.com/joachimvh/SPARQLAlgebra.js/pull/98 to be merged first.

lib/util/Ordering.ts
```diff
@@ -6,48 +6,79 @@ import { TermTransformer } from '../transformers/TermTransformer';
...
+import * as Err from './Errors';
 import type { ISuperTypeProvider, SuperTypeCallback, TypeCache, GeneralSuperTypeDict } from './TypeHandling';
 import { getSuperTypeDict } from './TypeHandling';
 
 // Determine the relative numerical order of the two given terms.
 /**
  * @param enableExtendedXSDTypes System will behave like when this was true. @deprecated
  */
-export function orderTypes(termA: RDF.Term | undefined, termB: RDF.Term | undefined, isAscending: boolean,
+export function orderTypes(termA: RDF.Term | undefined, termB: RDF.Term | undefined,
```
jitsedesmet, line 17: `orderTypes` is an exported function, can we just change the behavior like this? (requires major release?)

jitsedesmet (commented):
Everything looks good to me. I have some minor 'hints' that might make the code more consistent with the rest, but it's honestly just nit-picking.

lib/util/Ordering.ts
rubensworks: Let's just do a major update then :-)

The `isAscending` option felt a bit weird to me, and complicated the implementation too much, so I removed it.

lib/functions/RegularFunctions.ts
```diff
@@ -185,6 +189,18 @@ const equality = {
...
+      [ 'quad', 'quad' ],
+      context => ([ left, right ]) => {
+        const op: RegularFunction = new RegularFunction(RegularOperator.EQUAL, equality);
+        return bool(
+          (<E.BooleanLiteral> op.apply([ (<Quad> left).subject, (<Quad> right).subject ], context)).coerceEBV() &&
+          (<E.BooleanLiteral> op.apply([ (<Quad> left).predicate, (<Quad> right).predicate ], context)).coerceEBV() &&
+          (<E.BooleanLiteral> op.apply([ (<Quad> left).object, (<Quad> right).object ], context)).coerceEBV(),
+        );
+      },
+      false,
```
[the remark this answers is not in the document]
jitsedesmet: The flag (when true = default value) makes sure that the current implementation throws an error when passed an invalid literal. (some functions don't want this, since they handle invalid literals themselves (see equality of terms))
I don't think this implementation will run into trouble so we can keep it as is.

Issue #178, opened 2023-06-21
Refactor out redundancies in RegularFunctions and Ordering


rubensworks:
Functions such as lessThan, greaterThan, ... contain many redundancies.
Is there a reason why we can not rewrite all of them in terms of the ordering logic in Ordering?

@jitsedesmet perhaps you have some insights into this?

jitsedesmet:
Good question, I can't think of a reason why this wouldn't work, seems like Ordering only requires things in the SharedContext and this is available in the Regularfunctions. I think we should try it.

jitsedesmet:
For [a Comunica issue](https://github.com/comunica/comunica/issues/1212#issuecomment-1562513421) I also wanted to rewrite some aggregators to use the Ordering syntax. In the same PR I will look into this issue?

rubensworks:
Maybe we can start with just #1212, and leave this issue here for a later date (as I expect it will require a significant effort).

PR #180, opened 2023-07-08
Aggregators Min and Max call Ordering file

jitsedesmet:
This PR aims to resolve https://github.com/comunica/comunica/issues/1212.
[According to the spec](https://www.w3.org/TR/sparql11-query/#defn_aggMin), Min and Max should make use of the [SPARQL ORDER BY](https://www.w3.org/TR/sparql11-query/#modOrderBy) ordering definition. I assume sparqlee implements this in `Ordering.ts` (Both the spec and the implementation are wild to me, so please correct me if this assumption is wrong). 

~The issue also makes me wonder whether there are problems regarding `date` (and similar types) within the other aggregators, since all aggregators call the `typedValue` of a literal.  Before https://github.com/comunica/sparqlee/pull/163 this was always some basic JS type (if I recall correctly), but now there are also object typedValues, these do not implement the `+` operator for example.~ The other aggregators do call the regularFunctions so should work.

### DIFF
--- a/lib/util/Ordering.ts
+++ b/lib/util/Ordering.ts
@@ -94,10 +94,6 @@ function orderLiteralTypes(litA: RDF.Literal, litB: RDF.Literal,
   const myLitB = termTransformer.transformLiteral(litB);
 
   try {
-    // TODO: Calling apply does come at a small cost, searching for the correct overload...
-    //  @Ruben, we can fix this in an ugly way by using if statements (like it was before).
-    //  If we than also add an option to this function that will disable the fallback,
-    //  we could use the this function in our regular functions.
     if ((<E.BooleanLiteral> isEqual.apply([ myLitA, myLitB ], context)).typedValue) {
       return 0;
     }
--- a/lib/util/Ordering.ts
+++ b/lib/util/Ordering.ts
@@ -77,7 +77,7 @@ export function orderTypes(termA: RDF.Term | undefined, termB: RDF.Term | undefi
 
 function orderLiteralTypes(litA: RDF.Literal, litB: RDF.Literal,
   typeDiscoveryCallback?: SuperTypeCallback, typeCache?: TypeCache): -1 | 0 | 1 {
-  const isLess = regularFunctions[C.RegularOperator.LT];
+  const isGreater = regularFunctions[C.RegularOperator.GT];
   const isEqual = regularFunctions[C.RegularOperator.EQUAL];
   const context = {
     now: new Date(),
@@ -97,10 +97,10 @@ function orderLiteralTypes(litA: RDF.Literal, litB: RDF.Literal,
     if ((<E.BooleanLiteral> isEqual.apply([ myLitA, myLitB ], context)).typedValue) {
       return 0;
     }
-    if ((<E.BooleanLiteral> isLess.apply([ myLitA, myLitB ], context)).typedValue) {
-      return -1;
+    if ((<E.BooleanLiteral> isGreater.apply([ myLitA, myLitB ], context)).typedValue) {
+      return 1;
     }
-    return 1;
+    return -1;
   } catch {
     // Fallback to string-based comparison
     const compareType = comparePrimitives(myLitA.dataType, myLitB.dataType);
--- a/lib/util/Ordering.ts
+++ b/lib/util/Ordering.ts
@@ -1,16 +1,14 @@
 import type * as RDF from '@rdfjs/types';
 import * as LRUCache from 'lru-cache';
-import type { LangStringLiteral } from '../expressions';
-import { isNonLexicalLiteral } from '../expressions';
+import type * as E from '../expressions';
+import { regularFunctions } from '../functions';
 import { TermTransformer } from '../transformers/TermTransformer';
-import { TypeAlias, TypeURL } from './Consts';
-import type { ITimeZoneRepresentation } from './DateTimeHelpers';
-import { toUTCDate } from './DateTimeHelpers';
+import * as C from './Consts';
 import * as Err from './Errors';
-import type { ISuperTypeProvider, SuperTypeCallback, TypeCache, GeneralSuperTypeDict } from './TypeHandling';
-import { getSuperTypeDict } from './TypeHandling';
+import type { SuperTypeCallback, TypeCache } from './TypeHandling';
 
 // Determine the relative numerical order of the two given terms.
+// In accordance with https://www.w3.org/TR/sparql11-query/#modOrderBy
 /**
  * @param enableExtendedXSDTypes System will behave like when this was true. @deprecated
  */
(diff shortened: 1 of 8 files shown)


lib/util/Ordering.ts
```diff
@@ -79,49 +77,42 @@ export function orderTypes(termA: RDF.Term | undefined, termB: RDF.Term | undefi
...
-      }
-      return comparePrimitives((<LangStringLiteral>myLitA).language, (<LangStringLiteral>myLitB).language);
+    return 1;
+  } catch {
+    // Fallback to string-based comparison
+    const compareType = comparePrimitives(myLitA.dataType, myLitB.dataType);
+    if (compareType !== 0) {
+      return compareType;
     }
+    return comparePrimitives(myLitA.str(), myLitB.str());
```
jitsedesmet, line 114: If we add an extra parameter to the function that makes the catch rethrow instead of executing the fallback, that could allow us to easily fix https://github.com/comunica/sparqlee/issues/178

test/integration/util/Ordering.test.ts
```diff
@@ -103,7 +103,8 @@ describe('terms order', () => {
   it('dateTime type comparison', () => {
     genericOrderTestLower(dateTime('2000-01-01T00:00:00Z'), dateTime('2001-01-01T00:00:00Z'));
   });
-  it('langString type comparison', () => {
+  it.skip('langString type comparison', () => {
```
jitsedesmet, line 106: Code relates to https://github.com/comunica/sparqlee/blob/master/lib/util/Ordering.ts#L110-L115. @rubensworks do you recall why this case exists? The spec does not say anything about this I think. Would we like to have a langstring comparison? Should we add an `Operator Extension` for it?

jitsedesmet:
This seems to resolve the problem.
Query:
```sparql
SELECT (MIN(?of) as ?min) (MAX(?of) as ?max)
where {
    ?p ?o ?of.
}
```
over:
```n-triples
<http://example.com/apple> <http://www.w3.org/2001/XMLSchema#datePred> "2010-06-21+08:00"^^<http://www.w3.org/2001/XMLSchema#date>.
<http://example.com/pear> <http://www.w3.org/2001/XMLSchema#datePred> "2010-06-21-08:00"^^<http://www.w3.org/2001/XMLSchema#date>.
```
gives:
```
[
{"max":"\"2010-06-21-08:00\"^^http://www.w3.org/2001/XMLSchema#date","min":"\"2010-06-21+08:00\"^^http://www.w3.org/2001/XMLSchema#date"}
]
```

test/integration/util/Ordering.test.ts
rubensworks: Indeed, I think we added support for it according to SPARQL's operator extensibility: https://www.w3.org/TR/sparql11-query/#operatorExtensibility
We probably want to keep this in whatever form it can be implemented most easily.

rubensworks (approved):
Other than the single comment, this looks good to me!

test/integration/util/Ordering.test.ts
jitsedesmet: Makes sense, but if this is the case, we would need to define it in the regular function, so the inequality functions in FILTER also work accordingly. Do you agree?

rubensworks: Sure, but wouldn't the ineq already work?
The table [here](https://www.w3.org/TR/sparql11-query/#OperatorMapping) seems to indicate that inequality should work for all RDF terms (so that would include langStrings).

lib/util/Ordering.ts
rubensworks: The same `strict` parameter as the function above you mean?

test/integration/util/Ordering.test.ts
jitsedesmet: With inequality (ongelijkheid? Maybe I translated too literally?), I meant defining the `<` operator for it. I do not see that in the table?

lib/util/Ordering.ts
```diff
@@ -79,49 +77,38 @@ export function orderTypes(termA: RDF.Term | undefined, termB: RDF.Term | undefi
...
+    if ((<E.BooleanLiteral> isLess.apply([ myLitA, myLitB ], context)).typedValue) {
+      return -1;
     }
-    if (TypeURL.RDF_LANG_STRING in superTypeDictA && TypeURL.RDF_LANG_STRING in superTypeDictB) {
-      const compareType = comparePrimitives(myLitA.typedValue, myLitB.typedValue);
-      if (compareType !== 0) {
-        return compareType;
-      }
-      return comparePrimitives((<LangStringLiteral>myLitA).language, (<LangStringLiteral>myLitB).language);
+    return 1;
```
jitsedesmet, line 103: This else case might be false. While implementing https://github.com/comunica/sparqlee/pull/181 I saw that there are cases where `A < B` = False = `A > B` = false = `A < B`. In ordering, it might not be too bad(?) since in this case, we just want consistent ordering? However, this does matter regarding our discussing in https://github.com/comunica/sparqlee/pull/181

Example of such a case is `A = B = NaN`: https://www.w3.org/TR/xpath-functions/#func-numeric-less-than

test/integration/util/Ordering.test.ts
rubensworks: Ah, the smaller-than operator you mean, I thought you meant `!=`.
But yes, I think we want to keep it.

lib/util/Ordering.ts
rubensworks: I don't get what the problem is, can you elaborate?

jitsedesmet: Implemented here is a fall through chain:
1. A == B ?
2. A  <  B ?
3. A  >  B

What I mean is that the conclusion in step 3 is "false", since there are cases where non of the above hold. (I think 2 times the value NaN is such a case.) (I'd need to check again, this command is quite old)

rubensworks: If this is only the case for NaN, then this shouldn't be applicable, since NaN can't occur in SPARQL AFAIK.
So maybe it's a problem that can never occur?

jitsedesmet: I checked; It seems like [float](https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#float) does have "NaN" in its value space. More particular, the spec states:

> This datatype differs from that of [[IEEE 754-1985]](https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#ieee754) in that there is only one NaN and only one zero. This makes the equality and ordering of values in the data space differ from that of [[IEEE 754-1985]](https://www.w3.org/TR/2004/REC-xmlschema-2-20041028/#ieee754) only in that for schema purposes NaN = NaN.

However, in JavaScript, (NaN === NaN) = true, if I'm not mistaken, in current sparqlee implementation, we 
 do thus give a wrong answer.

PR #181, opened 2023-07-08
Refactor away redundency in inequality functions

jitsedesmet:
This PR aims to resolve https://github.com/comunica/sparqlee/issues/178 (working together but not dependent on each other is: https://github.com/comunica/sparqlee/pull/180)

### DIFF
--- a/lib/functions/RegularFunctions.ts
+++ b/lib/functions/RegularFunctions.ts
@@ -32,6 +32,7 @@ import type { IOverloadedDefinition } from './Core';
 import { RegularFunction } from './Core';
 import { bool, decimal, declare, double, integer, langString, string } from './Helpers';
 import * as X from './XPathFunctions';
+import { regularFunctions } from '.';
 
 const DF = new DataFactory<RDF.BaseQuad>();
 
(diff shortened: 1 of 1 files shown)


rubensworks (commented):
While this already simplifies the code significantly, my question in #178 was not really about writing these functions in terms of each other, but if it would be possible to only depend on the `orderTypes` function. Do you think this makes sense? Or would this cause issues that I'm not thinking of?

jitsedesmet:
@rubensworks given the implementation in https://github.com/comunica/sparqlee/pull/180 is what you want for `Ordering`, I would keep making the regular functions call the regular functions since it is more self-contained.
Also, if we would want to call ordering, we would not want the fall through case as mentioned in [a commend I made (and later resolved in favor of this PR)](https://github.com/comunica/sparqlee/pull/180#discussion_r1257265191)

rubensworks:
> Also, if we would want to call ordering, we would not want the fall through case as mentioned in https://github.com/comunica/sparqlee/pull/180#discussion_r1257265191

Ok, then we can change that I guess? That's the same reason why I added the `strict` parameter on `orderTypes` in d1e01c353c9c2ae810cece7ec4ff209ec1a57dd6.

jitsedesmet:
But why would you want this? I feel like this case is way better? You have all the logic of the regularfunctions in this one file? 
Why would you want to go: `RegularFunction -> Ordering -> RegularFunction` when you can do `RegularFunction -> RegularFunction`?

rubensworks:
> But why would you want this? I feel like this case is way better? You have all the logic of the regularfunctions in this one file?
Why would you want to go: RegularFunction -> Ordering -> RegularFunction when you can do RegularFunction -> RegularFunction?

Not sure I understand you correctly.
What I suggest is to just delegate to the `orderTypes` function. AFAIK, `orderTypes` has no external references back to regular functions.

rubensworks (approved):
Let's wait with merging this one until #180 is merged.

rubensworks:
Closes comunica/sparqlee#178

